PR #280, opened 2014-12-15
new example: Darcy flow in multiple compartments

vlukes:


vlukes:
The code is now cleaned up.

PR #282, opened 2015-01-26
Doc: Add Anaconda install notes

BubuLK:
Installation notes update:
- add Anaconda distribution info
- update tested versions
- bux-fixes
- minor reorganization and clenup


rc:
Thanks! It looks ok. Could you also fix the `download`_ and `sfepy download`_ links in doc/links.inc to http://sfepy.org/doc-devel/downloads.html? I have just noticed that they still point to the old site.

And (just a nit-picking), the commit message could be more like the PR message:

```
docs: update installation notes

- add Anaconda distribution info
- update tested versions
- bux-fixes
- minor reorganization and clenup
```

I will then merge it. :)

PR #286, opened 2015-03-22
Add boilerplate in eval_ns_forms.py

gladk:


rc:
Thanks! It might be worth doing the same with script/eval_tl_forms.py (and adding x to the file mode). Then all files in script/ would be executable scripts.

gladk:
Just to let you know. SfePy will soon be added to
Debian main archive. It hangs now in NEW-queue to
be checked by ftp-masters [1]. So users of Deb/Ubuntu
and other derivatives will be able to install it just by
"sudo apt-get install python-sfepy".

Cheers

[1] https://ftp-master.debian.org/new/sfepy_2015.1-1.html

Anton
 <https://ftp-master.debian.org/new/sfepy_2015.1-1.html>

rc:
Great news, thanks for telling! So is the change I proposed above worth doing (as part of this PR)?

gladk:
Yes, sure one can do it.

You are placing some files into the usr/bin, which are at the moment
absent in deb-package. If you think, they can be useful for users,
it would be good to add a prefix to all of them (for example sfepy_)
not to mix them with other thousands of files in usr/bin. And one more
important thing is the manpage for each file in usr/bin.

Best regards


Anton

2015-03-22 10:32 GMT+01:00 Robert Cimrman <notifications@github.com>:

rc:
I see, thank you for the info. The man pages should be doable with sphinx.

I am mostly using sfepy in-place, without installing it, so the installation has been a sort of a neglected sibling. There are too many scripts with too generic names etc. I am aware of that. We will have a look at other scientific python packages and think about this issue. Any further recommendations will be appreciated!

Best regards!
r.

gladk:
Let`s wait till the package will be accepted by our FTP-master, test it and
then one can go forward.

Cheers


Anton

2015-03-22 12:49 GMT+01:00 Robert Cimrman <notifications@github.com>:

rc:
BTW. where should a man page for a python package go to?

gladk:
Usually to /usr/share/man/man1
But it does not matter whether it is a python package or not.

PR #288, opened 2015-04-07
new mesh function: expand2d - converts tri/quad mesh to tetra/hexa

vlukes:


rc:
Thanks! Could you provide a test for this function? test_mesh_generators.py would seem a suitable location, or create a new file (test_mesh_ops.py?).

Maybe a better name would be extrude2d()? This is how this operation is called in CADs, IMHO.

A comment was added also to the commit.

vlukes:
On 2015-04-09 15:19, Robert Cimrman wrote:
> Thanks! Could you provide a test for this function?
> test_mesh_generators.py would seem a suitable location, or create a new
> file (test_mesh_ops.py?).
>
> Maybe a better name would be extrude2d()?

I was inspired by MSC.Mentat, it calls this function "expand".

>This is how this operation is called in CADs, IMHO.

Yes, but this is not the CAD function. The extrusion operation is 
usually driven by a 3D path. In my case it is just simple expansion to 3D.

rc:
Then the name is fine...

vlukes:
docstring fixed

rc:
Thanks! add a test, and I will merge it.

vlukes:
The test is done.

PR #289, opened 2015-04-24
allow matrix_hook to handle residual vector

vlukes:


rc:
OK! Just add it also to LCBCEvaluator.eval_residual() with mode 'lcbc_residual'(?).

vlukes:
LCBCEvaluator.eval_residual() calls BasicEvaluator.eval_residual() so the matrix_hook is processed.

rc:
Yes, but the two get a different vector - before and after LCBC application- see .eval_tangent_matrix().

vlukes:
Done, LCBCEvaluator.eval_residual() calls matrix_hook with mode 'lcbc_residual' now.

PR #295, opened 2015-06-25
remove dets_fast, use standard numpy linalg.det function instead

vlukes:
dets_fast function uses lapack_lite.dgetrf which has been removed in the new numpy version.


vlukes:
Instead of using lapack_lite in dets_fast use numpy.linalg.det for numpy version >= 1.8.

PR #301, opened 2015-10-15
lib.lapack bug fix

RexFuzzle:
As per http://docs.scipy.org/doc/scipy/reference/release.0.12.0.html#scipy-lib-lapack lib.lapack is depreciated and so I updated to the new linalg.

### DIFF
--- a/sfepy/solvers/eigen.py
+++ b/sfepy/solvers/eigen.py
@@ -121,7 +121,11 @@ class ScipySGEigenvalueSolver(EigenvalueSolver):
     def __init__(self, conf, **kwargs):
         EigenvalueSolver.__init__(self, conf, **kwargs)
 
-        import scipy.lib.lapack as llapack
+        try:
+            import scipy.linalg.lapack as llapack
+        except ImportError:
+            import scipy.lib.lapack as llapack
+
         self.llapack = llapack
 
     @standard_call


rc:
Thank you!

In order to make it work with older SciPy versions, could you make that conditional, based on ``scipy.__version__``?

RexFuzzle:
Cool, will get on it next week.

rc:
OK, thanks!

RexFuzzle:
So it looks like scipy hasn't been imported before this point and to get the version you have to import scipy- seems a bit unnecesary to import scipy just to get the version. I can do a try else instead?

rc:
OK, that would do: try the new version, and on exception revert to the old one.

rc:
Looks good. Only, please, squash the three commits to a single one (message something like "fix lib.lapack import"). I will then merge it. Thank you!

PS: The easiest way to do that is IMHO to reset the branch back and create a new single commit with the try-except method.

rc:
Final nit-picking: "Bug lib.lapack.bug fix" -> "lib.lapack import bug fix" :)

rc:
Sorry, I commented to the now-removed commit, so you probably missed that: It is recommended by Python gurus to catch a specific exception rather than all, so here, use `except ImportError:`, please. (I apologize for being pesky.)

rc:
Thank you, it is merged.

PR #306, opened 2016-02-16
fix gen_mesh_from_geom(), remove gen_mesh_from_poly()

vlukes:
typo: has been fixed also in the code :-) and the branch name is now fix-gen_mesh_from_geom.

polyfilename: removed as the argument, used tempfile.mkdtemp() for the temporary files.

There are many other things that should be improved in the code, maybe later.


rc:
Thanks!

PR #307, opened 2016-02-17
new and updated tests of gen_mesh_*()  functions

vlukes:


vlukes:
Does gen_extended_block_mesh() work correctly? When visualised only the "central block" is displayed.

PR #310, opened 2016-03-09
update meshio.py: support for msh file format (gmsh) - reading

vlukes:


rc:
Thank you!

Could you add a small example mesh to 'meshes/various_formats/' and update 'tests/test_meshio.py'?

vlukes:
The new test meshes (MSH) are there.

rc:
When viewing the 3D mesh with

```
./postproc.py --wire meshes/various_formats/msh_tetra.msh
```

The edges between domains seem doubled - is the mesh connected correctly?

vlukes:
Yes, there are doubled vertices at the interface of two material groups, it is just a mesh for testing purpose. Does it matter?

rc:
I would prefer a mesh that would work for a simulation.

vlukes:
OK, I pushed the new one.

rc:
The new one seems to be in VTK format, not gmsh?

vlukes:
Sorry, my fault, corrected.

rc:
Thanks you, the mesh is OK now.

I have just noticed that the both calls to `_read_bounding_box()` are wrong - they do not obey the values of `ret_fd`, `ret_dim` arguments to `MeshIO.read_bounding_box()`.

vlukes:
Calls to _read_bounding_box() - fixed.

rc:
Thanks!

PR #316, opened 2016-04-08
new: parallel computation of homogenized coefficients

vlukes:


rc:
Thank you, it looks like a nice contribution!

First, two minor things: ``coef_names`` and ``results`` variables in ``HomogenizationEngine.call()`` are unused and could be removed? There is also an empty line 215.

Then some tests in ``test_install.py`` fail, namely:

```
python ./phonon.py examples/phononic/band_gaps.py
```

fails, probably due to changed output of the homogenization engine - this should be easy to fix (in the test). Then

```
python ./phonon.py examples/phononic/band_gaps.py --phase-velocity
```

hangs indefinitely.

Then there is something strange as

```
./run_tests.py tests/test_homogenization_perfusion.py
```

fails (almost immediately - no computations seem to be done), while

```
./run_tests.py tests/test_homogenization_perfusion.py --debug
```

runs ok.

Can you reproduce the problems?

vlukes:
> First, two minor things: |coef_names| and |results| variables in
> |HomogenizationEngine.call()| are unused and could be removed? There is
> also an empty line 215.

removed

> |python ./phonon.py examples/phononic/band_gaps.py |
>
> fails, probably due to changed output of the homogenization engine -
> this should be easy to fix (in the test).

output lines are now in different order - fixed in the test

> |python ./phonon.py examples/phononic/band_gaps.py --phase-velocity |
>
> hangs indefinitely.

I had to disable multiprocessing in this case. There is a object 
(function) in shared arguments that can not be pickled - required by 
multiprocessing.

> Then there is something strange as
>
> |./run_tests.py tests/test_homogenization_perfusion.py |
>
> fails (almost immediately - no computations seem to be done), while
>
> |./run_tests.py tests/test_homogenization_perfusion.py --debug |
>
> runs ok.

I use stdout.flush() to ensure the correct order of lines in the output 
- solved by adding flush() method to OutputFilter class.

rc:
On 04/12/2016 10:09 AM, Vladimir Lukes wrote:
>> First, two minor things: |coef_names| and |results| variables in
>> |HomogenizationEngine.call()| are unused and could be removed? There is
>> also an empty line 215.
>
> removed
>
>> |python ./phonon.py examples/phononic/band_gaps.py |
>>
>> fails, probably due to changed output of the homogenization engine -
>> this should be easy to fix (in the test).
>
> output lines are now in different order - fixed in the test

OK, thanks!

As for having 'python ./homogen.py examples/homogenization/perfusion_micro.py' 
in test_install.py, IMHO it was added so that the micro is always tested - the 
regular test does not run it, when the coefficients are already computed and 
stored in the output directory.

A correct way would be for the tests to force recalculation of the 
coefficients/requirements in the test mode, even if the file is present. Could 
you have a look at it? It belong to another PR, though :)

>> |python ./phonon.py examples/phononic/band_gaps.py --phase-velocity |
>>
>> hangs indefinitely.
>
> I had to disable multiprocessing in this case. There is a object
> (function) in shared arguments that can not be pickled - required by
> multiprocessing.

OK. Or we can change the code so that a function is not passed around, if it is 
not complicated.

>> Then there is something strange as
>>
>> |./run_tests.py tests/test_homogenization_perfusion.py |
>>
>> fails (almost immediately - no computations seem to be done), while
>>
>> |./run_tests.py tests/test_homogenization_perfusion.py --debug |
>>
>> runs ok.
>
> I use stdout.flush() to ensure the correct order of lines in the output
> - solved by adding flush() method to OutputFilter class.
>

Good catch.

Thanks!
r.

vlukes:
> As for having 'python ./homogen.py
> examples/homogenization/perfusion_micro.py'
> in test_install.py, IMHO it was added so that the micro is always tested
> - the
> regular test does not run it, when the coefficients are already computed
> and
> stored in the output directory.

'perfusion_micro.py' is now tested by checking presence of some lines in 
the output string.

> A correct way would be for the tests to force recalculation of the
> coefficients/requirements in the test mode, even if the file is present.
> Could
> you have a look at it? It belong to another PR, though :)

OK, later.

VL

rc:
Thanks!

PR #319, opened 2016-04-19
fix arg_shapes in SDDotVolumeTerm

vlukes:


rc:
Thanks! The ``'parameter_mesh_velocity' : 'D'`` in the second set should not be needed, as the omitted values are taken from previous sets. Could you try that?

Also a detail - my editor indents the second line so that the ``'`` are aligned :)

vlukes:
OK, fixed.

rc:
Thanks!

Issue #164, opened 2012-01-30
Python 3 compatibility


rc:
As our dependencies are commencing to work in Python 3 (e.g. both NumPy and SciPy do), we should start thinking about it as well. I would like to have a single code base, that work for pythons 2.6(?) - 3.x (is this possible without too much hassle?).

Comment here, please, if you have some experience/thoughts in this area.
<p>Migrated from http://code.google.com/p/sfepy/issues/detail?id=162</p><hr/><h2>earlier comments</h2>
<p><strong>rashed.golam said, at 2011-12-21T12:52:06.000Z:</strong></p>
<p>we should make a dependencies list and see whats its status on porting to python 3.</p><p><strong>robert.cimrman said, at 2011-12-22T11:22:45.000Z:</strong></p>
<p>Yes - a new wiki page here (PortingToPython3) might work well for this purpose.</p>

takluyver:
Hi all, I offered on the Scipy user mailing list to help with the porting effort.

The first thing is to make sure that all the dependencies support Python 3. I went through the list, and pysparse, which looks like a dead project, appears to be Python 2 only. I haven't looked at what sfepy does with it, but would it be possible to drop the dependency in favour of something maintained like scipy.sparse?

It's also not clear whether [igakit](https://bitbucket.org/dalcinl/igakit/src) supports Python 3. It still seems to be actively maintained however. Checking this would be a good first step for someone to do.

After that, a lot of the changes will be straightforward, and can be done by running [python-modernize](http://python-modernize.readthedocs.io/en/latest/) over the code base. This takes care of things like `print` -> `print()` and `except ValueError, e` -> `except ValueError as e`. Python-modernize is based on 2to3, but keeps the code Python 3 compatible. It's up to you how you want to commit these changes - I might run one fixer at a time and commit after each, but you could do it as one big change, or break it up by chunks of the project.

Finally, there will be some changes that python-modernize can't handle. Hopefully these won't be too numerous - I just ran `python-modernize` on the whole lot, and with <10 manual changes on top of that it's gone from all tests failing to only half of them.

rc:
Wow that was quick, thank you! Either me or @BubuLK will look at this ASAP.

Dependencies:

- pysparse is used only for its (very good) eigenvalue problem solvers - it is an optional dependency. We use scipy.sparse for the sparse matrices and everything else.

- igakit is also only an optional dependency, so it is not critical to have it at the beginning. We can ping the author later.

[...]

PR #323, opened 2016-05-03
Rename extractor to extractor_sfepy

gladk:
extractor.py conflicts with the package python-extractor in Debian.
See [#822435](https://bugs.debian.org/822435)


rc:
Thank you!

Another possibility, which might be better in future (as also other sfepy scripts have rather generic names), is to install all the main scripts (see ``main_scripts`` in ``setup.py``) as data files only, and leave sfepy-run as a sole script to be installed into dist-packages/.

What do you think? Would that solve https://bugs.debian.org/822435?

gladk:
Sure, it will.

The only problem now is the name conflict in dist-packages so we "workarounded" it by adding
suffix. If extractor.py will be moved out of dist-packages in future sfepy releases, we will just
drop this workaround.

rc:
I can schedule that for 2016.2, to be out probably next week.

rc:
The https://github.com/rc/sfepy/commits/no-symlinks branch (https://github.com/rc/sfepy/commit/530ebc91af6efa04881e0fe38185595fe4d655ab) should do the trick - the main scripts are no longer installed into dist-packages/. (It does also other things, that need to be tested on windows, but that is not relevant here.) Could you verify that?

BTW. there is now a manpage for sfepy - this might be relevant for packaging. It can be built in the doc/ directory using ``make man`` and then viewed using ``man _build/man/sfepy-run.1``

gladk:
Thanks for fixing that. It works just fine.

Is it possible to generate manpage during the sphinx build? 
Also not a relevant to this issue question. Are there any plans to support Python 3?

Regards

rc:
OK, thanks for the testing.

Python 3 is planned (already in #164). Hopefully, @BubuLK will have some spare time to do that in not-so-far future.

What command do you use for building the sphinx docs? 'man' is just another sphinx build target such as 'html' or 'pdf'. It could be made so that e.g. ``python setup.py htmldocs`` built also the man page.

rc:
The 2016.2 release contains the fix, so let us close this.

Issue #328, opened 2016-05-24
Memory allocation error due to 32-bit integer overflow


cpetrich:
It appears, an overflowing 32-bit integer is being cast to 64 bit during memory allocation. 
For example, with an 80x80x80 block mesh I get in sfepy.discrete.Problem.time_update():

> sfepy: matrix shape: (11983194, 11983194)
> sfepy: assembling matrix graph...
> sfepy/discrete/common/extmods, sfepy/discrete/common/extmods/meshutils.c, mesh_graph, 298: error allocating 18446744065588459344 bytes (current: 670083736).
> mem_alloc_mem(): error exit!

18446744065588459344==0xfffffffe1bf1f750. Looks like int32 should be cast to size_t before multiplication.
Observed on Ubuntu 14.04 64bit with sfepy 2016.2.

rc:
Thanks for the report, I will check that. There might be similar issues at other places, so the fix might take some time.

rc:
What kind of problem was that? With a modified examples/diffusion/cube.py (80x80x80 hexahedral mesh, third order approximation) I am getting:

```
sfepy: matrix shape: (13144256, 13144256)
sfepy: assembling matrix graph...
sfepy: ...done in 103.75 s
sfepy: matrix structural nonzeros: 1609840448 (9.32e-06% fill)
```
Could you make a small example that fails, and attach it here (without the mesh :))?

cpetrich:
Thanks for following up. The following code fails on me on both Windows and Linux during ```pb.time_update()```:

```
import numpy
from sfepy.mesh import mesh_generators
from sfepy.discrete import (FieldVariable, Material, Integral, Function,
                            Functions,Integrals,
                            Equation, Equations, Problem)
from sfepy.discrete.fem import FEDomain, Field
from sfepy.mechanics.matcoefs import stiffness_from_youngpoisson
from sfepy.terms import Term
from sfepy.discrete.conditions import Conditions, EssentialBC
from sfepy.solvers.ls import ScipyIterative
from sfepy.solvers.nls import Newton

mesh=mesh_generators.gen_block_mesh(
        numpy.array([2,2,2]),
        numpy.array([80,80,80]),
        numpy.array([0,0,0]))
domain = FEDomain('domain',mesh)
omega = domain.create_region('Omega','all')
gamma_fix = domain.create_region('Gamma_fix','vertices in z == -1','facet')
fix = EssentialBC('fix', gamma_fix, {'u.all':0})

order = 2
[...]

PR #329, opened 2016-05-25
Fix mesh_graph() int32 overflow

rc:
A fix for #328. (work in progress)


rc:
While the number of matrix rows/columns and thus also column indices fit into 32 bit range without problems, the number of matrix nonzeros can easily exceed that. This means that also ``indptr`` array of the scipy CSR matrix needs to be 64 bit.

The following needs to be updated (or a 64 bit version created):

- ``mesh_graph()``
- matrix assembling function in ``assemble.pyx``

rc:
`mesh_graph()` was removed.

PR #331, opened 2016-06-15
Some work towards Python 3 support

takluyver:
As discussed in #164 

There are a lot of files changed, but most of the changes are simple automatic fixes made by `python-modernize`. If you want to review them, I suggest breaking it down by commits. Commit messages like "Run ... fixer" are automated changes, the others are manual changes I made to fix test failures.

This isn't complete Python 3 support, but it's hopefully a good start, and I figured that this PR was big enough already.

Everything should still work on Python 2, although I haven't actually checked this. There was a bit of confusion with the Variables and Problem classes in `sfepy.discrete`, but hopefully I've fixed it now.


run_tests.py
```diff
@@ -202,8 +202,9 @@ def main():
     else:
         op.walk(options.test_dir, run_tests, stats)
 
-    print '%d test file(s) executed in %.2f s, %d failure(s) of %d test(s)'\
-          % (stats[0], stats[3], stats[1], stats[2])
+    print('%d test file(s) executed in %.2f s, %d failure(s) of %d test(s)'
+          % (stats[0], stats[3], stats[1], stats[2]))
+
```
rc: Empty line addition not needed.

sfepy/discrete/problem.py
```diff
@@ -1142,7 +1142,7 @@ def create_evaluable(self, expression, try_equations=True, auto_init=False,
         """
         from sfepy.discrete.equations import get_expression_arg_names
 
-        variables = get_default(var_dict, {})
+        variables = Variables(six.itervalues(get_default(var_dict, {})))
```
rc: This was supposed to be a dict, not Variables - will check that.

rc:
Amazing, thanks! You have beaten me utterly to do that.

I have briefly looked at the commits:
- all the automatic ones seems ok
- I will check the manual ones (a few notes above - may be wrong)

Great work!

rc:
I am now trying to fix the remaining issues, see https://github.com/rc/sfepy/tree/py3. @takluyver do you have any tips on how to best install the python3 versions of dependencies alongside the python 2 versions? https://github.com/yyuu/pyenv? (I am using kubuntu.)

takluyver:
If you're happy with the Python versions in Ubuntu, you can install `python3-` packages with apt, e.g. `python3-numpy`, and with `python3-pip` installed, you'll have a `pip3` command to install further packages from PyPI.

If you prefer to switch to a separate environment to test on Python 3, I'd recommend Anaconda/Miniconda (Anaconda installs a load of packages to start with, Miniconda starts with a small base and you can install packages). They both contain the `conda` command, which lets you create environments and install packages into them. Pyenv can create environments in a similar way, but installing extension modules into them may require compiling.

rc:
Thanks, I will stick with the ubuntu packages for now.

BTW:

```
95 test file(s) executed in 44.21 s, 43 failure(s) of 154 test(s)
```

... looks already quite well!

takluyver:
:-) great! Feel free to ping me if there's anything that it's not clear how to fix.

rc:
I am now at

```
95 test file(s) executed in 68.55 s, 12 failure(s) of 155 test(s)
```

The remaining problems are caused mostly by the fact that python3 version of sympy seems not to be packaged in kubuntu 14.04. So I will try anaconda for this later.

Apart from that, there are some string/bytes issues related to pytables in HDF5MeshIO (e.g. line 1315 - writing a string ``mesh.name``) - any ideas how to fix that?

takluyver:
I think `pip3 install --user sympy` should work to install sympy. The `--user` flag installs it in a hidden directory under your home dir, so you don't need to sudo it.

With the Pytables issue, you mean [this line](https://github.com/sfepy/sfepy/blob/39343c3bf85f4011f9d84e74ee87e4440e1242ca/sfepy/discrete/fem/meshio.py#L1315), right? I think I ran into that as well and didn't see an obvious solution. What's the error it's giving?

rc:
Thanks for the sympy tips (no at 10 failures :)).

No, it is the line 1308 in that version of the file. The errors reads:

```
 File "/usr/lib/python3/dist-packages/tables/_past.py", line 35, in oldfunc
    return obj(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/tables/file.py", line 1144, in create_array
    flavor = flavor_of(obj)
  File "/usr/lib/python3/dist-packages/tables/flavor.py", line 196, in flavor_of
    "supported objects are: %s" % (type_name, supported_descs))
TypeError: objects of type ``str`` are not supported in this context, sorry; supported objects are: NumPy array, record or scalar; homogeneous list or tuple, integer, float, complex or bytes
```

and I am passing in ``mesh.name``, which is (in my current py3 master) ``<class 'str'>``

rc:
Also, on https://github.com/sfepy/sfepy/blob/39343c3bf85f4011f9d84e74ee87e4440e1242ca/sfepy/homogenization/micmac.py#L82, ``val`` is bytes, and I need to convert it to ``str`` - is using ``str(val, 'ascii')`` portable?

takluyver:
Aha, that makes sense. str in Py3 is unicode, and it sounds like pytables doesn't know how to store it natively.

You can encode a str to bytes (which pytables can store) with:

```python
mesh.name.encode('utf-8')
```

You may need a corresponding `.decode('utf-8')` in the code that loads it to get a str back.

You'll also need to work out whether the mesh.name API:

1. Always expects unicode on both Python versions
2. Always expects the native str type (so unicode on Py3, bytes on Py2)
3. Always expects str on Py3, but accepts either str or bytes on Py2.
4. Accepts both types on both versions.

You can make a case for any of those - I'd normally go for 1 or 3 for a new API, but it depends what you already accept on Py2.

takluyver:
`str(val, 'ascii')` doesn't work on Python 2. The general way to convert bytes to unicode is `val.decode(encoding)`. `encoding` may be 'ascii', if you know val can never contain any byte > 127 - otherwise, it gets more complex. If you control both the encoding and the decoding, 'utf-8' allows you store any character. If you're decoding bytes that may have been encoded by something else, 'latin1' will always decode *something*, but it may be wrong. Text is hard.

`.decode()` also assumes that what you start with is always bytes. On Py2, unicode sometimes turns up where you're expecting str (==bytes on Py2). If that's a possibility, then you can do:

```
if isinstance(val, bytes):
    return val.decode(...)
else:
    return val
```

rc:
Thanks for the information, I will try to think this over.

takluyver:
I should also say that working out the unicode/bytes distinction is usually the hardest bit of Python 3 porting, so the fact that you're down to 10 failures before running into this is pretty encouraging :-)

rc:
Good to know that ;)

rc:
```
95 test file(s) executed in 81.46 s, 3 failure(s) of 157 test(s)
```

This is about as far as I can get - one error is due to missing VTK, and two due to missing igakit.
So IMHO, after updating the docs/requirements, the PR will be ready to merge, so that it is tested in the wild.

Note: I did ``pip3 install --user mayavi`` and it failed because of missing VTK, but installed some packages that allowed more tests to pass. ``pip3 install --user sympy`` worked fine, other required packages are regular ubuntu 14.04 packages.

Could someone (ping @vlukes :))  try running the tests as well?

takluyver:
VTK 7 is compatible with Python 3 - I don't know what it takes to install that, but it is available.

vlukes:
95 test file(s) executed in 136.11 s, 3 failure(s) of 157 test(s)

The failures are related to missing igakit and VTK library.

rc:
Yes, I have just managed to install vtk 7 manually. And it turned out that igakit does support Python 3, so:

```
95 test file(s) executed in 86.95 s, 0 failure(s) of 157 test(s)
```

As concerns mayavi2, it fails to run, so postproc.py does not work either, but that is a separate issue (and using mayavi 4.4 with sfepy has other problems, cf. https://github.com/enthought/mayavi/issues/380).

rc:
Thanks Vladimir!

rc:
The docs are updated now - after someone checks that, I will do the merge :) Thanks to all who participated and especially to @takluyver who got us going! (@takluyver, it might be good if you fetch my changes from https://github.com/rc/sfepy/tree/py3, fast-forward your py3 branch, and push it, so that all is recorded here.)

takluyver:
Awesome! I've fetched your changes and pushed them to my branch, so they're part of this PR now. Let me know if you push any more and want me to update it.

rc:
Thanks! I will push the merge in a few minutes, if nobody protests :)

PR #334, opened 2016-08-24
fix docstring of SDDotVolumeTerm, SDDivTerm, SDDivGradTerm, SDConvect…

vlukes:
fix docstring of SDDotVolumeTerm, SDDivTerm, SDDivGradTerm, SDConvectTerm, SDGradDivStabilizationTerm, SDDiffusionTerm

### DIFF
--- a/sfepy/terms/terms_diffusion.py
+++ b/sfepy/terms/terms_diffusion.py
@@ -92,27 +92,29 @@ class SDDiffusionTerm(Term):
         - material:    :math:`K_{ij}`
         - parameter_q: :math:`q`
         - parameter_p: :math:`p`
-        - parameter_v: :math:`\ul{\Vcal}`
+        - parameter_mesh_velocity: :math:`\ul{\Vcal}`
     """
     name = 'd_sd_diffusion'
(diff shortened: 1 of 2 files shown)


sfepy/terms/terms_diffusion.py
```diff
@@ -92,7 +92,7 @@ class SDDiffusionTerm(Term):
         - material:    :math:`K_{ij}`
         - parameter_q: :math:`q`
         - parameter_p: :math:`p`
-        - parameter_v: :math:`\ul{\Vcal}`
+        - parameter_mesh_velocity: :math:`\ul{\Vcal}`
```
rc: Why this renaming? `parameter_v` seems to be correct according to `arg_types` below.

rc:
Thanks, seems OK, except the remark above.

vlukes:
The rest of sensitivity terms have "parameter_mesh_velocity" instead of "parameter_v".

rc:
I see - then rename it everywhere in the class, please, so that the docstring is consistent with the code.

vlukes:
I renamed it according to other sensitivity terms.

rc:
Thanks1

Issue #281, opened 2015-01-22
use argparse instead of optparse


rc:
As optparse is deprecated since Python version 2.7, it should be replaced by argparse (a part of Python standard library since 2.7).

sukhbinder:
Hi RC, 

With kids and family visiting grandparents, I have some time and can take up this task during the summer. Let me know where to start. 

Thanks

rc:
Great, thanks! It concerns the scripts in script/ and the top-level ones. It should be pretty straightforward, just follow [1]. Also, I will upgrade some of the scripts as an example.

[1] https://docs.python.org/2/library/argparse.html#upgrading-optparse-code

sukhbinder:
Thanks RC. Thanks for the link. I am aware of it and will be using this. Maybe I will begin with the postproc.py script.

I will need your help in the github requests as I am new to it. Thanks

rc:
Sure! Feel free to ask on our ML, or here.

heczis:
Are there any tests for this? Should they be written at all?

rc:
There are no tests, besides the ones in test_install.py.

I do not think we need tests for this - just a manual verification that each option works after the change as it used to work before should be enough. The only "tricky" options are the one with custom parsers and/or choices.

PR #336, opened 2016-09-02
Use argparse instead of optparse #281

heczis:
A fix for #281.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -24,7 +24,7 @@ import sys
 import time
 import os
 import os.path as op
-from optparse import OptionParser
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
 
 import sfepy
 from sfepy.base.conf import ProblemConf, get_standard_keywords
@@ -148,12 +148,10 @@ def get_dir(default):
         out = op.normpath(op.join(sfepy.data_dir, default))
     return out
 
-usage = """%prog [options] [test_filename[ test_filename ...]]"""
-
 help = {
-    'dir' : 'directory with tests [default: %default]',
+    'dir' : 'directory with tests [default: %(default)s]',
     'out_dir' : 'directory for storing test results and temporary files'
-    ' [default: %default]',
+    ' [default: %(default)s]',
     'debug' : 'raise silenced exceptions to see what was wrong',
     'filter-none' : 'do not filter any messages',
     'filter-less' : 'filter output (suppress all except test messages)',
@@ -162,31 +160,37 @@ help = {
 }
 
 def main():
-    parser = OptionParser(usage=usage, version="%prog " + sfepy.__version__)
-    parser.add_option("", "--print-doc",
-                      action="store_true", dest="print_doc",
-                      default=False, help=help['print-doc'])
-    parser.add_option("-d", "--dir", metavar='directory',
-                      action="store", dest="test_dir",
-                      default=get_dir('tests'),
-                      help=help['dir'])
-    parser.add_option("-o", "--output", metavar='directory',
-                      action="store", dest="out_dir",
-                      default=get_dir('output-tests'),
-                      help=help['out_dir'])
-    parser.add_option("", "--debug",
-                      action="store_true", dest="debug",
-                      default=False, help=help['debug'])
-    parser.add_option("", "--filter-none",
-                      action="store_true", dest="filter_none",
-                      default=False, help=help['filter-none'])
-    parser.add_option("", "--filter-less",
-                      action="store_true", dest="filter_less",
-                      default=False, help=help['filter-less'])
-    parser.add_option("", "--filter-more",
-                      action="store_true", dest="filter_more",
-                      default=False, help=help['filter-more'])
-    options, args = parser.parse_args()
+    parser = ArgumentParser(description=__doc__,
+                            formatter_class=RawDescriptionHelpFormatter
+    )
+    parser.add_argument("-v", "--version", action="version",
+                        version="%(prog)s " + sfepy.__version__)
+    parser.add_argument("--print-doc",
+                        action="store_true", dest="print_doc",
+                        default=False, help=help['print-doc'])
+    parser.add_argument("-d", "--dir", metavar='directory',
+                        action="store", dest="test_dir",
+                        default=get_dir('tests'),
+                        help=help['dir'])
+    parser.add_argument("-o", "--output", metavar='directory',
+                        action="store", dest="out_dir",
+                        default=get_dir('output-tests'),
+                        help=help['out_dir'])
+    parser.add_argument("--debug",
+                        action="store_true", dest="debug",
+                        default=False, help=help['debug'])
+    parser.add_argument("--filter-none",
+                        action="store_true", dest="filter_none",
+                        default=False, help=help['filter-none'])
+    parser.add_argument("--filter-less",
+                        action="store_true", dest="filter_less",
+                        default=False, help=help['filter-less'])
+    parser.add_argument("--filter-more",
+                        action="store_true", dest="filter_more",
+                        default=False, help=help['filter-more'])
+    parser.add_argument("test_filename", nargs="*", default=[])
+
+    options = parser.parse_args()
 
     if options.print_doc:
         print(__doc__)
@@ -195,11 +199,10 @@ def main():
     run_tests = wrap_run_tests(options)
     stats = [0, 0, 0, 0.0]
 
-    if len(args) >= 1:
-        for test_filename in args:
+    if len(options.test_filename) > 0:
+        for test_filename in options.test_filename:
             dirname, filename = op.split(test_filename)
             run_tests(stats, dirname, [filename])
-
     else:
         for dirpath, dirnames, filenames in os.walk(options.test_dir):
             run_tests(stats, dirpath, filenames)
--- a/script/extract_surface.py
+++ b/script/extract_surface.py
@@ -14,7 +14,7 @@ from __future__ import absolute_import
 import sys
 from six.moves import range
 sys.path.append('.')
-from optparse import OptionParser
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
 
 import numpy as nm
 import scipy.sparse as sp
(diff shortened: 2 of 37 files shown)


rc:
Thanks, good work!

All tests pass in Python 2.7, but travis-ci reports some problems with Python 3.4

Overall it looks good, but there are some minor issues that need to be addressed (besides the 3.4 compatibility):

- limit the line lengths to 80 (79) characters
- the merge commit does nothing and should be removed (I can help with that)
- check that all scripts run also without any arguments (e.g. postproc.py fails)
- argparse was already used in some of the examples (e.g. the parallel ones) and I would prefer using the same importing style (``from argparse import ArgumentParser``)
- also leave ``options`` instead of ``args`` - no need to change that (we use ``options`` as a parameter name of many functions)

That is all I see now at the first sight. :)

Thanks again!

heczis:
I guess the corrections are finished now.
The failing test in python3 was caused by using the `version` kwarg in the call of ArgumentParser constructor.
I tried to run all the changed scripts without any arguments, but no problems emerged. Could it be that you tested postproc.py using python3? It would probably fail due to the above mentioned python3 incompatibility.
Also, thanks for the comments and guidance ;-)

script/plot_condition_numbers.py
```diff
@@ -109,7 +110,7 @@ def main():
         tt = time.clock()
         mtx, iels = term.evaluate(mode='weak', diff_var='u')
         output('...done in %.2f s' % (time.clock() - tt))
-        mtx = mtx[0][0, 0]
+        mtx = mtx[0,0]#[0][0, 0]
```
rc: Thanks for fixing that - you can remove the commented-out code.

script/extract_surface.py
```diff
@@ -81,29 +81,30 @@ def surface_components(gr_s, surf_faces):
...
+                        action="store_true", dest="no_surface",
+                        default=False,
+                        help="do not output surface [default: %(default)s]")
+    parser.add_argument('filename_in')
+    parser.add_argument('filename_out')
+    options = parser.parse_args()
+
+    filename_in = options.filename_in
+    filename_out = options.filename_out
+
```
rc: Superfluous line.

rc:
Thanks, overall it looks good!

The ``postproc.py`` problem is still there:

```
user@th203-p03-ntc projects/sfepy-git((d5a0901...))$ python2.7 ./postproc.py 
Traceback (most recent call last):
  File "./postproc.py", line 102, in <module>
    (' where #n is the number of plots in the specified direction '
TypeError: 'str' object is not callable
```
- it is caused by superfluous ``()`` at the lines 102-103.

After correcting that, I get:

```
user@th203-p03-ntc projects/sfepy-git((d5a0901...))$ python2.7 ./postproc.py
Traceback (most recent call last):
  File "./postproc.py", line 506, in <module>
    main()
  File "./postproc.py", line 496, in main
    view = view_file(filenames, filter_names, options)
  File "./postproc.py", line 281, in view_file
    domain_specific=options.domain_specific)
  File "/home/eldaran/projects/sfepy-git/sfepy/postprocess/viewer.py", line 915, in call_mlab
    offscreen=self.offscreen)
  File "/home/eldaran/projects/sfepy-git/sfepy/postprocess/sources.py", line 38, in create_file_source
    fmt = os.path.splitext(filename[0])[1]
IndexError: list index out of range
```

- the length of ``options.filenames`` should be checked, and help should be printed if it is zero.

heczis:
Here it is. The only thing I doubt is checking the number of positional arguments.
`argparse` has a way of doing this: set the `nargs` for the argument in question; If the user does not give enough arguments, print `usage` followed by a message of what was violated. E.g.:
```
$ examples/linear_elasticity/linear_viscoelastic.py
sfepy: fading memory length: 6
usage: linear_viscoelastic.py <output file in HDF5 format>
linear_viscoelastic.py: error: too few arguments
```
The printed usage-message is, however, usually missing detailed description of the options - this can be due to overriding the default value of `usage`. The automatic message (without setting the `usage`) would be in the above case:
```
$ examples/linear_elasticity/linear_viscoelastic.py
sfepy: fading memory length: 6
usage: linear_viscoelastic.py [-h] [-n ii] [-e ii] OUTPUT_FILE
linear_viscoelastic.py: error: too few arguments
```
Any way, to get the detailed description of options, the user has to ask for help explicitly, using the `-h` option. I find that OK; we could add e.g. "Use ./%(prog)s -h to get detailed help." to `usage`.

rc:
Thanks, ``postproc.py`` looks ok now.

As for the number of arguments checks, maybe we could just set ``nargs=*``, and check for correct length? Something like:

```
 diff --git a/examples/linear_elasticity/linear_viscoelastic.py b/examples/linear
index 31e7f08..128eb13 100644
--- a/examples/linear_elasticity/linear_viscoelastic.py
+++ b/examples/linear_elasticity/linear_viscoelastic.py
@@ -270,7 +270,7 @@ def main():
     }
 
     parser = ArgumentParser(usage=usage)
-    parser.add_argument(metavar='OUTPUT_FILE', dest='output_file',
+    parser.add_argument(metavar='OUTPUT_FILE', dest='output_file', nargs='*',
                         help='output file in HDF5 format')
     parser.add_argument('-n', '--node', type=int, metavar='ii',
                         action='store', dest='node',
@@ -280,7 +280,12 @@ def main():
                         default=299, help=msgs['element'])
     options = parser.parse_args()
 
-    filename = options.output_file
+    if len(options.output_file) == 1:
+        filename = options.output_file[0]
+
+    else:
+        parser.print_help()
+        sys.exit(1)
 
     tensor_names = ['cauchy_strain',
```

heczis:
Yes, we could do that - it is the original way these things were written.

What I am suggesting is to take advantage of `argparse` doing the checks
for us. It might be more work to do at this point and it might change the
interface of the scripts slightly, but I do believe it is a good way to
implement the argument parsing, since it uses the default mechanism offered
by argparse, thereby - hopefully - enforcing consistency.

Also, `argparse` can handle mutually exclusive options and groups of
options and checks this, too. I used this feature in simple.py.

rc:
OK, that sounds reasonable - would you like to add it to this PR? Or create a new one after this is merged. Anyway - adding "Use ./%(prog)s -h to get detailed help." to ``usage`` and using ``nargs`` where appropriate seems good to me.

heczis:
Right now, I would add "Use ./%(prog)s -h to get detailed help." to `usage` and use `nargs` and merge. I'll push the final version in the afternoon ;-)

heczis:
Well, I think it's ready now.

run_tests.py
```diff
@@ -148,12 +148,13 @@ def get_dir(default):
         out = op.normpath(op.join(sfepy.data_dir, default))
     return out
 
-usage = """%prog [options] [test_filename[ test_filename ...]]"""
+usage = """%(prog)s [options] [test_filename[ test_filename ...]]
+Use ./%(prog)s -h to get detailed help."""
```
rc: It is not necessary here - the script runs all the tests without arguments, and -h prints this info anyway.

rc:
Thanks! Just "Use ./%(prog)s -h to get detailed help." is not needed in ``run_tests.py`` and also it might be better to use 'Use "python %(prog)s -h" to get detailed help.' so that it is relevant also for systems where you cannot run python script just so (there are such...) :) It is almost ready to merge!

heczis:
I see one more way to change the value of `usage`: use the default value. Unlike `optparse`, `argparse` composes it from the actual options, e.g. (BTW this is an example of why I included the "Use..." message in run_tests.py, too):
```
$ ./run_tests.py --verbose
usage: run_tests.py [-h] [-v] [--print-doc] [-d directory] [-o directory]
                    [--debug] [--filter-none] [--filter-less] [--filter-more]
                    [test_filename [test_filename ...]]
run_tests.py: error: unrecognized arguments: --verbose
```
This way we could ommit the "Use ... to get detailed help" message hoping that the user would guess that the `-h` option prints help. We can still add some text to this, but my feeling is, that any further explanation should be part of the `description`, which is printed as part of the help message (the default help message consists of: `usage`, `description`, help on each argument, and possibly an `epilog`).

Why I didn't suggest this earlier:
* I was unaware that `optparse` uses a default value that is independent of the options expected.
* I saw overriding the default value in almost all scripts, theregore I thought it was an intention. (Was it?)
* I consulted the format of usage with a colleague of mine offering two options:
 - `<prog> [options] POS_ARG_1 POS_ARG_2`
 - `<prog> [-h] [-v] [-o directory] POS_ARG_1 POS_ARG_2`

  and she chose the first one as being more readable. Since then I asked more people and the second variant also gained some support, which changed my mind as well.

My suggestions of how to conclude this PR:
* Merge now, just after I change './%(prog)s' to 'python %(prog)s'; with or without the "Use..." message in run_tests.py - I personally believe it should be there.
* Revise the overriden `usage` messages, replace them with the default ones, and either:
 - Add any extra explanation, where it was previously (postproc.py and many others)
 - Move any extra explanation from `usage` to `description`

What do you think? Maybe we could hire a UX designer, do some tests on a group of users, and then decide :-)

rc:
Thanks for giving it a thought. I guess having the default usage is ok - ``-h`` is wide-spread enough for people to recognize it as the help option. Then the extra texts that are now in usage could be moved to description. So simply renaming ``usage`` to ``description``,  and removing superfluous text (like 'usage:...') would do, right?

The ``description`` could almost always be defined simply as ``__doc__.rstrip()`` -

rc:
Thanks for giving it a thought. I guess having the default usage is ok - ``-h`` is wide-spread enough for people to recognize it as the help option. Then the extra texts that are now in usage could be moved to description. So simply renaming ``usage`` to ``description``,  and removing superfluous text (like 'usage:...') would do, right?

The ``description`` could almost always be defined simply as ``__doc__.rstrip()`` -

heczis:
Here it is.
The only problem was with probe.py: I would split some arguments in mutually exclusive groups, which `argparse` does not support. In the end, I just left the code the way it was and added the two use cases to the description.

rc:
Great, all looks ok, just remove the now-useless ``usage`` variables from

```
examples/linear_elasticity/shell10x_cantilever_interactive.py
script/extract_surface.py
```

Thanks!

heczis:
Oops, sorry for that. Both should be fixed by now.

rc:
Thanks, it is merged now!

PR #337, opened 2016-09-08
run_tests.py: display the test number to get a better view

vlukes:


rc:
Thanks! I like the idea.

The number seems to be displayed only with ``--debug`` and ``--filter-none``. What about passing the test file number into ``run_test()`` and printing it at the line 79? Also ``TestCommon.report()`` could support that.

BTW. the number you print is a test file number, not a test number - maybe ``test a:b`` could be used, where ``a`` is a test file number and ``b`` the number of a test within that file.

vlukes:
The idea is to show the test file numbers and test numbers only  in a debug mode (--debug or --filter-none). Without these numbers it is difficult to orient in the outputs where are many lines.

Now:
 - test file numbers are displayed next to the filenames
 - lines are labelled as [x]
 - if there are multiple tests in the test file, lines are labelled as [x/y]

rc:
Thanks for the update. What about putting the test number in front of the test file name instead of behind? IMHO that would make it visually easier to spot the numbers.

rc:
OK for me, but there is an unwanted interaction with 'ls.cm_pb' solver - try: ``./run_tests.py --filter-none tests/test_input_vibro_acoustic3d.py``. Would it be easy to fix that? If not, I will merge as it is...

vlukes:
What about this solution?

rc:
Great!

rc:
Great!

rc:
Great!

PR #345, opened 2016-10-21
Fix pytables use to 3.x API

turjani:


rc:
Thanks for tackling this issue! We need to have this fixed also for #341 :)

Do you have a pytables version with the new API to run the tests against? IMHO you need to edit all pytables objects - for example:

```
fd = pt.open_file(filename, mode='w', title='Recursive dict dump')
....
fd.create_array(group, '_' + str(key), enc(val), '%s data' % key)
```
etc.

Note also, that adding one character might cause more than 80 chars per line, so you might need to add a new line at some places.

rc:
Fixed by #348 ...

PR #346, opened 2016-10-31
Fix setting field coordinates for higher order elements

vlukes:


rc:
Thanks! Do you have a test for this feature? Also, please, change ``extra_dof`` to ``extra_dofs`` to have the name consistent with other similar uses, e.g. ``FEField.remove_extra_dofs()``.

vlukes:
At the moment, I have no test for this feature. ``extra_dof`` fixed.

rc:
Well, I meant just some example, where you use that feature (that could be used as a test in future). Because it looks OK, but I have no way of verifying that it works - so, does it work for you?

BTW. the travis-ci fail above sometimes happens and is caused by ``tests/test_input_linear_elastic_mM.py`` - it fails when run for the first time, but it does not fail when re-run again with ``--raise``. I have a hunch that it is related to some race condition caused by saving/reading the homogenized coefficients or correctors.

Issue #342, opened 2016-10-10
update pytables use to 3.x API


rc:
PEP8 naming conventions etc. see http://www.pytables.org/MIGRATING_TO_3.x.html#major-api-changes. Influenced files:

```
sfepy/base/ioutils.py
sfepy/discrete/fem/meshio.py
sfepy/discrete/iga/io.py
sfepy/optimize/free_form_def.py
shaper.py
```

PR #348, opened 2016-11-01
fix pytables compatibility issue: renamed openFile() -> open_file()

vlukes:


rc:
Thanks! btw. there is also #345 (to fix #342) - are you aware of those?

vlukes:
Now I see, this request is equivalent to #345.

rc:
Yours seems more complete.

PR #351, opened 2016-11-09
homogenization engine clean up

rc:
@vlukes, could you check this branch? There might be other possibilities for clean up and/or adding more tests.


vlukes:
Now, it seems to be OK. Actually, I'm working on homogenization for large deformations, so I will probably revise the engine soon.

rc:
OK, I need to first fix the test to make the dependencies ordering unique... Then I will merge it.

PR #352, opened 2016-11-28
Update for multiscale simulations of  large deformations

vlukes:
main changes:
 - solving multiple micro-problems in "one step", micro configurations are updated using the macroscopic deformation gradient and correctors
 - the volumes required in computation of coefficients are now computed also as coefficients - can be easily parallelized
 - caching of periodic matches (deformed microstructures are no more periodic)
 - improved parallelization mechanism, new multiproc.py module
 - using global caches for periodic matches and (some) mappings

### DIFF
--- a/examples/homogenization/nonlinear_homogenization.py
+++ b/examples/homogenization/nonlinear_homogenization.py
@@ -1,0 +1,234 @@
+# -*- coding: utf-8 -*-
+import numpy as nm
+from sfepy.homogenization.utils import define_box_regions
+import sfepy.homogenization.coefs_base as cb
+import sfepy.discrete.fem.periodic as per
+from sfepy.base.base import Struct
+from sfepy.terms.terms_hyperelastic_ul import\
+    HyperElasticULFamilyData, NeoHookeanULTerm, BulkPenaltyULTerm
+from sfepy.terms.extmods.terms import sym2nonsym
+from sfepy.discrete.functions import ConstantFunctionByRegion
+from sfepy import data_dir
+import sfepy.linalg as la
+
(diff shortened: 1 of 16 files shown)


rc:
Great job! I have just a few remarks:

- [ ]  sfepy/discrete/fem/periodic.py needs a clean-up (spaces around function args)
- [ ] ``match_coors()``: ``coor1`` -> ``coors1``, ``which`` is undefined.
- [ ] I wonder if there is a better place to initialize ``domain.mesh.coors_act`` than ``set_mesh_coors()``?
- [ ] add into sfepy/base/multiproc.py:

```
managers; Process; # Make pyflakes happy.
```
- [ ] document sfepy/base/multiproc.py and the new functions in engine.py etc. (e.g. ``calculate_req()`` has tons of arguments)
- [ ] remove line 157 in engine.py
- [ ] unused imports (output, MiniAppBase, six), unused line 98 in sfepy/homogenization/homogen_app.py
- [ ] Why using capitals for VOLUME coefficients?  What about "'Volume'"?

vlukes:
Yes, there could be a better place to initialize 'domain.mesh.coors_act'. But the other solutions require some extra options or major changes in the code. Now, I'm not sure how to make it better.

All other remarks have been reflected in the commits.

rc:
OK, thanks! Is there a (simple) way of testing the large deformation homogenization besides running the full example? (Something in terms of ``tests/test_homogenization_engine.py``).

vlukes:
I have updated tests/test_homogenization_engine.py to test the functions 
related to splitting into chunks, merging chunks and volume computation. 
Could it be like this?

rc:
Seems good! Merging soon...

PR #354, opened 2016-12-07
New numpy version compatibility

lokik:
In new numpy a comparison with None is element-wise.


rc:
Thanks! Could you rebase on the current master, and split the line:

```
       if isinstance(subs, tuple) and subs[0] is None and subs[1] is None:
            subs = None
```

NIce catch.

rc:
I did it myself - merging in a while... :)

PR #355, opened 2016-12-08
Fix redefining the help keyword

heczis:
Couldn't call help(...) while in pdb.
Not sure if this is the preferred way, but suggesting anyway.


rc:
Yes, newer scripts already use ``helps`` instead of ``help`` - merging, thanks!

PR #360, opened 2017-01-02
Use the logging module for output in test_install.py

heczis:
I just thought... why not use `logging` for logging.
Things are set-up in one place at the beginning. Also, the format of `test_install.log` is changed slightly, but that was my intention, too (so that the failed command is displayed above the debug output).


rc:
Nice, thanks!

PR #362, opened 2017-01-24
Update splinebox

vlukes:


rc:
@vlukes, thanks! By any field you mean something other than the mesh coordinates? Do you have some example, where this feature is used?

vlukes:
Yes, It can be used to parametrize an arbitrary scalar or vector field in a given region.
Usage: I want to identify a material parameter, let's say alpha(x), which is defined piecewise constant on cells. Using a splinebox parametrization of alpha(x):
1) I can reduce the number of optimization parameters (ncells -->  number of splinebox control points)
2) resulting alpha(x) is "smooth", i.e. avoiding the checkerboard effect

rc:
@vlukes, thanks for explaining. I am merging now. BTW. it would be good to have an example of the material identification with this.

PR #365, opened 2017-02-23
fix saving surface regions

vlukes:


rc:
Thanks! Only use ``tdim`` (topological dimension) instead of ``dim`` (space (coordinates) dimension).

rc:
Thanks! Merging...

PR #372, opened 2017-03-14
fix changing of the material shape

vlukes:
Is it really necessary to change the shape of the material data?


rc:
Thanks, using ``.reshape()`` is better.

PR #373, opened 2017-03-16
new PiezoStrainTerm

vlukes:


rc:
Thanks! Just fix the docstring (superfluous '}' at the end of the definition, so that the docs can be built.
Also, could you incorporate it into ``post_process()`` in the `piezo_elasticity.py` example?
Otherwise LGTM.

vlukes:
The docstring is fixed and the new term is used in piezo-elasticity example.

rc:
Thanks, merging...

PR #375, opened 2017-03-18
Saving custom structured data to h5 file in problem.save_state

lokik:
We discussed it personally.

### DIFF
--- a/sfepy/base/ioutils.py
+++ b/sfepy/base/ioutils.py
@@ -119,18 +119,17 @@ def enc(string, encoding='utf-8'):
     """
     Encode given string or bytes using the specified encoding.
     """
-    val = string.encode(encoding)
-    return val
+    if sys.version_info > (3, 0):
+        string = string.encode(encoding)
+    return string
 
 def dec(val, encoding='utf-8'):
     """
     Decode given bytes using the specified encoding.
     """
-    if isinstance(val, bytes):
-        return val.decode(encoding)
-
-    else:
-        return val
+    if isinstance(val, bytes) and sys.version_info > (3, 0):
+        val = val.decode(encoding)
+    return val
 
 ##
 # 27.04.2006, c
--- a/sfepy/base/ioutils.py
+++ b/sfepy/base/ioutils.py
@@ -565,3 +565,22 @@ def bytes_from_str(string):
     if sys.version_info > (3, 0):
         return string.encode('utf8')
     return string
+
+class HDF5ContextManager:
+    def __init__(self, filename, *args, **kwargs):
+        self.filename = filename
+        self.file = None
+        self.args = args
+        self.kwargs = kwargs
+
+    def __enter__(self):
+        if isinstance(self.filename, pt.File):
+           return self.filename
+        else:
+           self.file = pt.open_file(self.filename, *self.args, **self.kwargs)
+           return self.file
+
+    def __exit__(self, type, value, traceback):
+        if self.file:
+           self.file.close()
+           self.file = None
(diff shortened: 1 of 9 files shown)


sfepy/base/ioutils.py
```diff
@@ -9,6 +9,9 @@
 import glob
 from .base import output, ordered_iteritems, Struct, basestr
 import six
+import pickle
+import numpy as np
```
rc: There is already ``import numpy as nm`` - use that in the new code, please.

sfepy/base/ioutils.py
```diff
@@ -386,3 +389,89 @@ def read_sparse_matrix_hdf5(filename, output_format=None):
         mtx.sort_indices()
 
     return mtx
+
+def save_to_pt(pt_file, group, data):
+    """ Save custom data to h5 file group to be restored by read_from_pt
+        allow saving lists, dicts, numpy arrays, scalars and all pickleable
+        objects
+    """
+
```
rc: Some docstring is better than none, but if you could adapt it to the [sfepy docstring standard](http://sfepy.org/doc-devel/developer_guide.html#docstring-standard), it would be even better.

sfepy/base/ioutils.py
```diff
@@ -386,3 +389,89 @@ def read_sparse_matrix_hdf5(filename, output_format=None):
...
+    elif isinstance(data, str):
+        save_value('str', np.array(data))
+
+    else:
+        save_value('pickle', np.array(pickle.dumps(data)))
+
+def read_from_pt(pt_file, group):
+    """ Read data from h5 file group saved by save_from_pt """
+    def load_list():
```
rc: BTW. using ``save_from_pt()`` might be better as it indicates  that the name refers to a function.

sfepy/discrete/fem/meshio.py
```diff
@@ -10,7 +10,7 @@
                              insert_static_method, output, get_default,
                              get_default_attr, Struct, basestr)
 from sfepy.base.ioutils \
-     import skip_read_line, read_token, read_array, read_list, pt, enc, dec
+     import skip_read_line, read_token, read_array, read_list, pt, enc, dec, read_from_pt, save_to_pt
```
rc: Split the line, please (max. 79 characters).

rc:
Thanks! Overall, it looks good, but see the comments above. Also, would you mind using ``read_from_hdf5()`` instead of ``read_from_pt()`` and ``write_to_hdf5()`` instead of ``save_to_pt()`` to be in a better agreement with the existing functions?

 Then the last thing: add a test for this functionality into `tests/test_meshio.py`.

lokik:
Comments was worked in, test and ability to save sparse matrix added.

lokik:
New bunch of commits:
- improvments according to your comments (hope I include them all)
- saving meshes and iga domain to hdf5
- HDF5ContextManager
- test

Please not that currently the new test failed: restoring of IGDomain.regions 
created during creating domains and stored into vertex_set_bcs cannot be
restored in the way we discussed, because there is a mapping function 
in creating process.
We can either
- commit this part of test and disable restoring of this stuff
- add a parameter to IGDomain.__init__ to disable remaping of regions to
 vertex_set_bcs
- something else?

There are also two (to storing to hdf5) unrelated improvments
- fix in problem.copy to correct copying of Problem subclasses
- ability to add functions to existing Problem

rc:
Thanks! I will check it soon. For now, remove, please, the merge commit (how did it happen?), and name the branch with some meaningful name, e.g. save-custom-data-to-hdf5, so that the eventual merge message does not contain just the generic name 'master'.

rc:
@lokik, could you remove the merge commit? I could then add to this branch more easily... Thanks!

rc:
@lokik, The updated code with some my modifications is now in 27e99c7128273fe99425379785c89dd4a16df382. I will merge/close this as soon as you tell me it works for you.

rc:
Merged alternate version.

PR #376, opened 2017-03-20
fix: avoid number truncation in region definitions

vlukes:


rc:
Thanks! I guess it might need more:
```
In [2]: '%e' % (1./3)
Out[2]: '3.333333e-01'
```
So, maybe:
```
In [3]: '%.16e' % (1./3)
Out[3]: '3.3333333333333331e-01'
```

Or, the number of digits could be related to the ``eps`` argument. What do you think?

vlukes:
I think '%.16e' is satisfactory.

rc:
OK, merging...

PR #380, opened 2017-03-28
Fix things to be compatible with both python 2 and 3

heczis:
Could not regenerate the documentation using `python3` due to changes in the `types` module.
Changed things according to [Dive into Python 3](http://www.diveintopython3.net/porting-code-to-python-3-with-2to3.html#types)


rc:
Thanks!

PR #384, opened 2017-03-31
fix the shape of the gradient array provided by 'ev_grad'

vlukes:
now: (n_el, n_qp, dim, n_c)

But internally, gradients still have shape (n_el, n_qp, n_c, dim)!

### DIFF
--- a/sfepy/terms/terms_navier_stokes.py
+++ b/sfepy/terms/terms_navier_stokes.py
@@ -316,7 +316,6 @@ class GradTerm(Term):
         vg, _ = self.get_mapping(parameter)
 
         grad = self.get(parameter, 'grad')
-        grad = nm.ascontiguousarray(grad.transpose((0, 1, 3, 2)))
 
         fmode = {'eval' : 0, 'el_avg' : 1, 'qp' : 2}.get(mode, 1)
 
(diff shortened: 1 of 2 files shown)


rc:
Now I recall, why the `ev_grad` result shape was `(n_el, n_qp, n_c, dim)` - it was to be consistent with `Variable.evaluate_at()` in the `'grad'`mode. So the shape was so on purpose - this is not a fix but a change in behaviour (*). IMHO it is a good change, to be consistent with other terms.

(*) BTW. this reverts 8d28b697d184ea2da84c3718a787c76f510ac12b, which, as I checked now, broke `script/save_basis.py`, and `examples/diffusion/sinbc.py` without anyone noticing. So, in fact, it is actually a fix - of something different, but call it 'change' instead of 'fix' in the commit and branch name anyway - there was nothing wrong with the actual shape per se.

We could change also `Variable.evaluate_at()` to return shape `(n_coor, dim, n_components)`, but that would mean  `(n_coor, 1, n_components)` in `'val'` mode, which might break many things. Also, `(n_coor, n_components, dim)` allows an addition of higher derivatives in future. So let's leave that as it is - what do you think?

vlukes:
I agree that changing behavior of `Variable.evaluate_at()` is not a good idea. But term `ev_grad` is aimed for data export and the exporting functions expect `(n_el, n_qp, dim, 1)`. So, I would distinguish between 'internal' functions like `evaluate_at()`, which can remain unchanged, and exporting/user oriented functionalities, that behave more  intuitively.

PR #385, opened 2017-03-31
change the shape of the gradient array provided by 'ev_grad'

vlukes:
now: (n_el, n_qp, dim, n_c)

But internally, gradients still have shape (n_el, n_qp, n_c, dim)!


rc:
See also #384 for discussion. Merging soon...

PR #393, opened 2017-04-20
Fix generators next

heczis:
Generators do not have the `next` method in python 3.
All found occurences replaced by calls to the built-in function, which is available since python 2.6.


rc:
Thanks, good catch!

Issue #382, opened 2017-03-29
fix installed build testing instructions


rc:
http://sfepy.org/doc-devel/installation.html#installed-build is wrong - `sfepy-run run_tests` should be used.

rc:
Fixed by  #397.

Issue #366, opened 2017-03-02
installation and download info: improve for conda install


rc:
It would be good to move the conda [installation instructions](http://sfepy.org/doc-devel/installation.html#multi-platform-distributions-notes) to a more visible place, and update the text. It should be also mentioned on the [downloads page](http://sfepy.org/doc-devel/downloads.html). The link to [conda-forge packages](https://anaconda.org/conda-forge/sfepy) should be added.

PR #397, opened 2017-05-09
Doc install

BubuLK:
Updated doc-install section:
- Issue #382 
- Issue #366 
- misc doc update/cleanup


doc/installation.rst
```diff
@@ -9,43 +9,45 @@ Installation
...
+
+*SfePy* is known to work on various flavors of Linux, Intel-based MacOS and
+Windows. It can work with both Python 2.7 and Python 3.x.
 
-*SfePy* is known to work on various flavors of Linux, on Intel Macs and
-Windows. It can work with both Python 2.7 and Python 3. Depending on Python
-installation and OS used, replacing ``python`` by ``python3`` might be required
-in all the commands below (e.g. in :ref:`compilation`) in order to use Python 3.
+Depending on Python installation and OS used, replacing ``python`` by ``python3`` might be required
+in all the commands below (e.g. in :ref:`compilation`) in order to use Python 3. However we strictly recommend to use *Python Virtual Environments* (see `Python virtualenv`_ or `Conda Environments`_) for these purposes.
```
rc: How `Python virtualenv`? helps here?

What about: "However, for new users, especially on Windows, we recommend to use ??? for these purposes."

doc/installation.rst
```diff
@@ -9,43 +9,45 @@ Installation
...
 
-The easiest way to install SfePy is to use `Anaconda`_, see `Multi-platform
-Distributions Notes`_.
+The far easiest way we recommend to use *SfePy* on any supported platform is to install `Anaconda`_ Python distribution
+(see `Multi-platform Python Distributions Notes`_) and follow `Generic SfePy Installation Instructions`_.
```
rc: I would just use:

The easiest way in install *SfePy* and its dependencies on any supported platform is to install `Anaconda`_ Python distribution, see `Multi-platform Python Distributions Notes`_.

The `Generic SfePy Installation Instructions`_ are not needed, when installing sfepy from conda-forge.

doc/installation.rst
```diff
@@ -54,26 +56,25 @@ Other dependencies:
 - To be able to (re)generate the documentation: `Sphinx`_, `numpydoc`_, LaTeX,
   see :ref:`how_to_regenerate_documentation`.
 - If `doxygen` is installed, the documentation of data structures and functions
-  can be automatically generated by running "``python setup.py
-  doxygendocs``". Then see "``doc/html/index.html``".
+  can be automatically generated by running ``python setup.py
+  doxygendocs``. Then see ``doc/html/index.html`` for results.
 - Mesh generation tools use `pexpect` and `gmsh` or `tetgen`.
```
rc: What about: Then open ``doc/html/index.html`` in your web browser.

doc/installation.rst
```diff
@@ -54,26 +56,25 @@ Other dependencies:
...
 On Linux, consult the package manager of your favorite distribution,
-see `Platform-specific Notes`_.
+see `Platform-specific Python Installation Notes`_.
 
-On Windows a completely free scientific-oriented distributions
-`Python(x,y)`_ and `WinPython`_ can be used, but some missing packages
-may need to be installed manually. Instructions for installing
-Python(x,y) can be found in `Windows Using Python(x,y)`_.
+On Windows we highly recommend a free version of commercials multi-platform scientific
+Python distributions `Anaconda`_ (from `Continuum Analytics`_) or `Enthought Canopy`_
```
rc: a free version of commercials -> free versions of commercial

doc/installation.rst
```diff
@@ -54,26 +56,25 @@ Other dependencies:
...
-Python distributions `Enthought Canopy`_ (formerly Enthought
-PythonDistribution) and `Anaconda`_ (`Continuum Analytics`_) can be
-used (and recomended), see `Multi-platform Distributions Notes`_.
+A completely free scientific-oriented distributions `WinPython`_ or `Python(x,y)`_
+can be used, but some compatibility issues or missing packages may expected (Instructions
+for installing `Python(x,y)`_ can be found in `Windows Using Python(x,y)`_).
 
-*SfePy* can be used without any installation by running the scripts from the
+*SfePy* can be used without any installation by running requested scripts from the
 top-level directory of the distribution, or can be installed locally or
```
rc: What about: requested scripts -> its main scripts and examples

rc:
Thanks, well done! I have some remarks/questions + your updates inspired me to several further recommendations/possible improvements:

- limit the line lengths in `installation.rst` to 79 chars, please
- at least on Linux, I prefer installing system python packages and using sfepy in-place without installation, so I would word the initial section differently, see the commit comments - why do you think a virtual environment is preferable in general?
- replace MacOS with OS X?
- instead of "Generic SfePy Installation Instructions", we could use "Installing SfePy from Sources"
- instead of "Platform-specific Python Installation Notes", we could use "Notes on Installing SfePy Dependencies on Various Platforms", and state, that those are "user-contributed instructions".
- for an in-place build, the ``python run_tests.py`` and ``python sfepy-run run_tests`` are equivalent and the former one command is shorter, so I would not remove that from the docs - we may mention both ways.
- for an installed build, does `python sfepy-run run_tests` work at some platform? It does not on Linux, as `sfepy-run` is not in the current directory - one must use just `sfepy-run run_tests`, as, on the other hand, `sfepy-run` is on the executable search `PATH`. How is this on Windows/OS X?
- remove instructions for old platforms: "Fedora 8" (2007), "Older Versions of Ubuntu"
- are Python(x,y) instructions still valid? If not, I would simply remove them.
- see also the commit comments

It is great that you started to address those issues, people keep stumbling on the related problems. Maybe we need a FAQ :)

doc/installation.rst
BubuLK: BTW: for system-based (package) install (especially for Linux), there are no such problems. But it's the only one case...
Please let's talk about it...?

rc: Yeah, but isn't that already a part of anaconda (which is described in anaconda docs)? Why mentioning Python virtualenv here explicitely?

BubuLK:
- limit the line lengths in installation.rst to 79 chars, please
 OK, done. What about CLI texts?
- replace MacOS with OS X? - No! OS X is obsolete, MacOS is now official Mac OS name. ;)
- are Python(x,y) instructions still valid? If not, I would simply remove them.
 AFAIK still valid, but obsolete... (2005, Python 2.7 only)

More comments follow...

doc/installation.rst
rc: Yes, I would just mention anaconda, so that it does not look that one must use a virtualenv. In practice, there is IMHO no mess - people have either Python 2.7, or 3.6. People having more versions are usually developers that know what they are doing, and do not need to be instructed.

doc/installation.rst
```diff
@@ -47,35 +56,37 @@ Python packages required for using *SfePy*:
...
   parts of primer/tutorial.
 
 On Linux, consult the package manager of your favorite distribution,
 see `Platform-specific Python Installation Notes`_.
 
-On Windows we highly recommend a free version of commercials multi-platform scientific
-Python distributions `Anaconda`_ (from `Continuum Analytics`_) or `Enthought Canopy`_
-(formerly Enthought Python Distribution). See `Multi-platform Python Distributions Notes`_
-for more details.
+On Windows we highly recommend a free versions of commercial multi-platform
```
rc: Remove 'a' in 'a free', please :)

doc/installation.rst
```diff
@@ -84,8 +95,8 @@ hold.
 
 .. _generic_install:
 
-Generic SfePy Installation Instructions
----------------------------------------
+Installing SfePy from sources
```
rc: sources -> Sources
Note that this change needs to be reflected in links like `Generic SfePy Installation Instructions`_...

rc:
OK, thanks for updates and explanation! The cli examples can be as long as needed - no 80 char limit there.

rc:
If possible, remove the merge commit, that is not needed. I can help with that.

rc:
Thanks, it is getting into a good shape!

rc:
Thanks, LGTM!

PR #409, opened 2017-06-27
Sphinx conf

BubuLK:
- corrected IPython console examples (tutorial.rst) to work with older IPython versions
- some under-hood Sphinx configuration changes:
  - removed custom ipython_console_highlighting.py and replaced with the standard one,
  - replaced deprecated pngmath with imgmath extension.


rc:
Thanks, removing files is always good ;-)

Could you rebase this on the master? Assuming `upstream` is the main sfepy repo:
```
git stash # Do this if you have uncommitted changes.)
git checkout master
git fetch upstream
git reset --hard upstream/master
git checkout Sphinx-conf
git rebase master
git stash pop # Only do this if you did `git stash`
```

Then, does
```
python setup.py pdfdocs
```
work for you? It looks like the latex_preamble is not used in the latex mode. The html mode seems ok, though.

rc:
Thanks, LTGM!

PR #410, opened 2017-06-30
new option to convert_mesh.py script: '-t' convert quad/hexa elements…

vlukes:
… to tri/tetra


script/convert_mesh.py
```diff
@@ -117,6 +121,26 @@ def main():
...
+
+        if conns is not None:
+            nelo = conns.shape[0]
+            output('initial mesh: %d elements' % nelo)
+
+            new_conns = elems_q2t(conns)
+            nn = new_conns.shape[0] // nelo
+            new_cgroups = nm.repeat(mesh.cmesh.cell_groups, nn)
+
+            output('... %d elements' % new_conns.shape[0])
```
rc: With output, `...` usually denotes un-indent one level - maybe better would be to have simply:
```
output('new mesh: %d elements' % new_conns.shape[0])
```

rc:
Thanks!

Issue #350, opened 2016-11-09
enable AppVeyor CI on github


rc:
AppVeyor.com is a Continuous Integration service to build and run tests under Windows.

PR #416, opened 2017-09-05
Deploment-CI

BubuLK:
Support to enable AppVeyor CI on github #350

- updated existing .travis.yml to switch to container-based build (faster)
- added basic AppVeyor config file
  - tests on x86/x86_64 platforms
  - TBD: register (and configure) main SfePy project on appveryor.com
- updated/synced CI testing configs:
  - tests for Python 2.7/3.6 only
  - don't run 'python run_tests.py --raise' (if tests failed) (unecessary test delay? )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,14 +56,10 @@ script:
 
   - python setup.py build_ext --inplace
   - python run_tests.py; export SFEPY_TEST_RESULTS=$?; (exit $SFEPY_TEST_RESULTS)
-
-#
-# Probably not needed here...
-#
-#  - |
-#    if [ $SFEPY_TEST_RESULTS -ne 0 ]; then
-#      python run_tests.py --raise
-#    fi
+  - |
+    if [ $SFEPY_TEST_RESULTS -ne 0 ]; then
+      python run_tests.py --raise
+    fi
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,20 @@
-# After changing this file, check it on: http://lint.travis-ci.org/
+#
+# SfePy minimal travis.yml config file
+#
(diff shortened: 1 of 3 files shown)


rc:
Great, thanks!

I guess I should enable AppVeyor for the sfepy repo, right?

.travis.yml
```diff
@@ -35,23 +40,30 @@ before_install:
...
-      python run_tests.py --raise
-    fi
+
+#
+# Probably not needed here...
+#
+#  - |
+#    if [ $SFEPY_TEST_RESULTS -ne 0 ]; then
+#      python run_tests.py --raise
+#    fi
```
rc: This is useful, because "python run_tests.py" does not provide any useful info on why a test failed. That is why on failure we re-run the tests without the silenced output. Maybe `--filter-none` instead of `--raise` might be better so that it does not stop on the first error.

Another possibility is to add `--filter-none` to the first call.

BubuLK:
Yes. In addition you should tweak GUI configuration (some settings are merged from GUI and config file... see https://www.appveyor.com/docs/build-configuration/#appveyoryml-and-ui-coexistence for further info).

.travis.yml
BubuLK: OK, I'll add --filter-none switch to first call...

P.S. My original idea was: you get pass/fail info on each test and you can re-run test localy.

rc: Well, you may not be able to run it locally, if you do not have that particular platform ;)

Actually, `--raise` is better - `--filter-none` does not show the exception. So I guess the original version is the most convenient.

rc:
Do you happen to know, why the travis tests now run >15 min. instead of about 6 min.? Might it be caused by setting the following?

```
sudo: false
dist: trusty
```

Issue #406, opened 2017-06-19
clean up cython/C sources to prevent compiler warnings


rc:
Too many warnings obscure useful information...

PR #423, opened 2017-10-23
C clean up

vlukes:
issue #406


rc:
Thanks!

BTW. @BubuLK has also a PR addressing #406 almost ready. Luckily, many things were done in an orthogonal way, so I will cherry-pick from both...

rc:
Merged.

PR #430, opened 2017-12-04
Update convert mesh

rc:
Hi @vlukes, could you review? I have updated the preprocessing doc a bit. Thank!


vlukes:
Doc seems to be all right. '--2d' feature is useful!

PR #431, opened 2017-12-04
new ev_surface_grad term

vlukes:


rc:
Thanks!

Most of the  code seems to be the same as in ev_grad - the only differences are the integration type and the term docstring - would it be possible to merge the two terms somehow?

vlukes:
Yes, of course, surface and volume terms  can be unified (it is done). That was also my original idea, first step was to create a new term, second step was to unify them. And as usual, the first one has been done and the second one has been forgotten.

vlukes:
I have also added ev_surface_div term.

rc:
Thanks, LGTM! Merging...

PR #432, opened 2017-12-04
update fields.evaluate_at() for evaluating complex fields

vlukes:


sfepy/discrete/common/fields.py
```diff
@@ -339,21 +340,34 @@ def evaluate_at(self, coors, source_vals, mode='val', strategy='general',
...
-                            dtype=source_vals.dtype)
+                            dtype=source_dtype)
             cmode = 1
 
         ctx = self.create_basis_context()
 
-        evaluate_in_rc(vals, ref_coors, cells, status, source_vals,
-                       self.get_econn('volume', self.region), cmode, ctx)
+        if source_vals.dtype in complex_types:
+            valsi = vals.copy()
```
rc: Why is this line needed?

rc:
Thanks! Could you add a test, please,  for this new functionality,  into ``tests/test_projections.py`` or ``tests/test_mesh_interp.py``?

sfepy/discrete/common/fields.py
rc: I see, I mistook that for source values.

vlukes:
There is a new test in ``tests/test_mesh_interp.py``.

rc:
LGTM! Is there a (simple) way of comparing the evaluated result also with the original data? Also, the omega variable is unused :)

vlukes:
I don't  see an easy way to do that. It was not the intention of this test. I think that ``evaluate_at`` for real values is tested enough elsewhere. Unused variable omega is removed.

PR #433, opened 2017-12-08
update mesh conversion: extract cells by material id

vlukes:


rc:
Thanks, useful addition!

Some remarks:
- Usually, the short options have a single letter - I suggest removing ``-pm``, this is not a very common operation.
- Use, please, ``--save-per-mat`` instead of ``--save_per_mat``
- Most os.path imports in sfepy are ``import os.path as op``
- The help message could be more verbose, something like:

    extract cells by material id and save them into a separate mesh file with a name based on `filename_out` and the id; preserves original mesh vertices

doc/preprocessing.rst
```diff
@@ -370,6 +370,15 @@ argument to `convert_mesh.py` script::
...
 
+In order to extract the cells by the physical groups use the conversion script
+with ``--save_per_mat`` (``-pm``) argument::
+
+  script/convert_mesh.py -pm screwdriver_full.vtk screwdriver.vtk
+
+It produces `screwdriver_matid_1.vtk` and `screwdriver_matid_2.vtk` files
+containing only the cells of a given physical group and all vertices of the
+original mesh.
+
```
rc: It saves those meshes in addition to the output mesh - maybe mention that?

vlukes: Yes, I added the comment to this.

vlukes:
The remarks are taken into account in the updated commit.

PR #434, opened 2017-12-12
Add reading of axisymmetric elements to AbaqusMeshIO

heczis:
I needed to read a mesh created in Abaqus that contained axisymmetric elements, which were not implemented. This code reads them as ordinary 2D elements (since SfePy does not take care of axial symmetry by the use of dedicated elements, AFAIK).


rc:
Thanks! Could you rebase your branch onto the current master?

heczis:
Oops, sorry tor that. Should be OK now.

PR #435, opened 2017-12-18
script/convert_mesh.py: new --remesh option

rc:
@vlukes could you check this PR? Thanks!


rc:
Added better help and volume remeshing support (without generating a surface mesh).

PR #436, opened 2017-12-18
new ProbeFromFile class: init VTK probe using a given file

vlukes:


rc:
Thanks! Could you add an example showing how to use it?

PR #443, opened 2018-01-25
Add new interactive hyperelastic example

heczis:
The example contains a few things that were not clear to me in the beginning, therefore I kindly ask others to check (apart from code quality) both correctness of the theory and its implementation. Thanks for any comments!

### DIFF
--- a/examples/large_deformation/hyperelastic_tl_up_interactive.py
+++ b/examples/large_deformation/hyperelastic_tl_up_interactive.py
@@ -9,7 +9,7 @@ given by
 
 where :math:`\overline I_1` and :math:`\overline I_2` are the first
 and second main invariants of the deviatoric part of the right
-Cauchy-Green deformation tensor :math:`\overline{\bf C}`. The coefficients
+Cauchy-Green deformation tensor :math:`\overline{\ull{C}}`. The coefficients
 :math:`C_{(10)}` and :math:`C_{(01)}` are material parameters.
 
 Components of the second Piola-Kirchhoff stress are in the case of an
@@ -20,8 +20,8 @@ incompressible material
 
 where :math:`p` is the hydrostatic pressure.
 
-Large deformation is described using the total Lagrangian formulation in this
-example. Incompressibility is treated by mixed displacement-pressure
+The large deformation is described using the total Lagrangian formulation in
+this example. The incompressibility is treated by mixed displacement-pressure
 formulation. The weak formulation is:
 Find the displacement field :math:`\ul{u}` and pressure field :math:`p`
 such that:
@@ -34,8 +34,8 @@ such that:
     \intl{\Omega\suz}{} q\, (J(\ul{u})-1) \difd{V} = 0
     \;, \quad \forall q \;.
 
-Following formula holds for axial true (Cauchy) stress in the case of uniaxial
-stress:
+The following formula holds for the axial true (Cauchy) stress in the case of
+uniaxial stress:
 
 .. math::
     \sigma(\lambda) =
@@ -47,16 +47,19 @@ where :math:`\lambda = l/l_0` is the prescribed stretch (:math:`l_0` and
 
 The boundary conditions are set so that a state of uniaxial stress is achieved,
 i.e. appropriate components of displacement are fixed on the "Left", "Bottom",
-and "Near" faces and monotonously increasing displacement is prescribed on the
-"Right" face. This prescribed displacement is then used to calculate
-:math:`\lambda` and to convert second Piola-Kirchhoff stress to true (Cauchy)
-stress.
+and "Near" faces and a monotonously increasing displacement is prescribed on
+the "Right" face. This prescribed displacement is then used to calculate
+:math:`\lambda` and to convert the second Piola-Kirchhoff stress to the true
+(Cauchy) stress.
 
-**Note**
+Note on material parameters
+---------------------------
 
 The relationship between material parameters used in the *SfePy* hyperelastic
-terms (:class:`NeoHookeanTLTerm <sfepy.terms.terms_hyperelastic_tl.NeoHookeanTLTerm>`,
-:class:`MooneyRivlinTLTerm <sfepy.terms.terms_hyperelastic_tl.MooneyRivlinTLTerm>`)
+terms (:class:`NeoHookeanTLTerm
+<sfepy.terms.terms_hyperelastic_tl.NeoHookeanTLTerm>`,
+:class:`MooneyRivlinTLTerm
+<sfepy.terms.terms_hyperelastic_tl.MooneyRivlinTLTerm>`)
 and the ones used in this example is:
 
 .. math::
@@ -64,7 +67,8 @@ and the ones used in this example is:
 
     \kappa = 2\, C_{(01)} \;.
 
-**Usage Examples**
+Usage Examples
+--------------
 
 Default options::
 
@@ -76,15 +80,28 @@ To show a comparison of stress against the analytic formula::
 
 Using different mesh fineness::
 
-  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py --shape 5 5 5
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py \
+    --shape 5 5 5
 
 Different dimensions of the computational domain::
 
-  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py --dims 2 1 3
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py \
+    --dims 2 1 3
 
 Different length of time interval and/or number of time steps::
 
-  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py -t 0,15,21
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py \
+    -t 0,15,21
+
+Use higher approximation order (the ``-t`` option to decrease the time step is
+required for convergence here)::
+
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py \
+    --order 2 -t 0,2,21
+
+Change material parameters::
+
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py -m 2 1
 """
 from __future__ import print_function, absolute_import
 import argparse
@@ -120,6 +137,21 @@ def get_displacement(ts, coors, bc=None, problem=None):
 def plot_graphs(
         material_parameters, global_stress, global_displacement,
         undeformed_lenght=1.0):
+    """
+    Plot a comparison of true stress computed by FEM and using an analytic
+    formula.
+
+    Parameters
+    ----------
+    material_parameters : list or tuple of float
+        C10 and C01 coefficients
+    global_displacement
+        Total displacement for each time step, from FEM
+    global_stress
+        True (Cauchy) stress for each time step, from FEM
+    undeformed_lenght : float
+        Length of undeformed specimen
+    """
     c10, c01 = material_parameters
 
     stretch = 1 + np.array(global_displacement) / undeformed_lenght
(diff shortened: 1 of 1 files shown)


rc:
Thanks, I will check it.

From a quick look, could you change the command line arguments to match other interactive examples?

- The  `gen_block_mesh()` could use `--dims`, `--shape` or even `--centre`.
- Use `-p`, `--plot`.

See  examples/multi_physics/biot_parallel_interactive.py.

For the time stepping, what about using something like `-t`, `--ts=t0,t1,n_step`?

heczis:
I changed the command line arguments. I usually like to let `argparse` handle list-like arguments, but in this case (with `--dims`, `--shape`, and `--centre` being lists of either floats or ints on one hand and `--ts` being a structure of two floats and one int on the other hand) it introduces a little inconsistency. Any thoughts on this?
I could probably also add some tests before merging, right? These could be written using `unittest` as a proof of concept. Or should I do this separately, perhaps with a discussion in `sfepy-devel` on what to test etc.?

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+
+    eq_balance = Equation('balance', term_neohook+term_mooney+term_pressure)
+    eq_volume = Equation('volume', term_volume_change-term_volume)
+    equations = Equations([eq_balance, eq_volume])
+
+    ### Solvers ###
+    ls = ScipyDirect({})
+    nls_status = IndexedStruct()
+    nls = Newton(
+        {'is_linear' : False},
```
rc: The default number of iterations is 1, so you need to increase it here - use, for example,  `'i_max' : 5`, and maybe also tweak the tolerances. On the other hand, `'is_linear' : False` is the default, and is not needed :)

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+def get_displacement(ts, coors, bc=None, problem=None):
+    """
+    Define time-dependent displacement
+    """
+    out = 1. * ts.time * coors[:, 0]
+    return out
+
+def plot_graphs(
+        material_parameters, global_stress, global_displacement,
+        undeformed_lenght):
```
rc: `undeformed_lenght` -> `undeformed_length`

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+    m = Material(
+        'm', mu=2*c10, kappa=2*c01,
+    )
+
+    ### Boundary conditions ###
+    x_sym = EssentialBC('x_sym', regions['Left'], {'u.0' : 0.0})
+    y_sym = EssentialBC('y_sym', regions['Near'], {'u.1' : 0.0})
+    z_sym = EssentialBC('z_sym', regions['Bottom'], {'u.2' : 0.0})
+    disp_fun = Function('disp_fun', get_displacement)
+    displacement = EssentialBC('displacement', regions['Right'], {'u.0' : disp_fun})
```
rc: The line is too long.

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+:math:`\lambda` and to convert second Piola-Kirchhoff stress to true (Cauchy)
+stress.
+
+Note on material parameters
+---------------------------
+
+The relationship between material parameters used in the *SfePy* hyperelastic
+terms (:class:`NeoHookeanTLTerm <sfepy.terms.terms_hyperelastic_tl.NeoHookeanTLTerm>`,
+:class:`MooneyRivlinTLTerm <sfepy.terms.terms_hyperelastic_tl.MooneyRivlinTLTerm>`)
+and the ones used in this example is:
```
rc: The lines are too long.

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+In this model, the deformation energy density per unit reference volume is
+given by
+
+.. math::
+    W = C_{(10)} \, \left( \overline I_1 - 3 \right)
+        + C_{(01)} \, \left( \overline I_2 - 3 \right) \;,
+
+where :math:`\overline I_1` and :math:`\overline I_2` are the first
+and second main invariants of the deviatoric part of the right
+Cauchy-Green deformation tensor :math:`\overline{\bf C}`. The coefficients
```
rc: Use `\ull{C}` to be consistent.

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+incompressible material
+
+.. math::
+    S_{ij} = 2 \, \pdiff{W}{C_{ij}} - p \, F^{-1}_{ik} \, F^{-T}_{kj} \;,
+
+where :math:`p` is the hydrostatic pressure.
+
+Large deformation is described using the total Lagrangian formulation in this
+example. Incompressibility is treated by mixed displacement-pressure
+formulation. The weak formulation is:
```
rc: The large deformation...   The incompressibility...

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+
+.. math::
+    \intl{\Omega\suz}{} \ull{S}\eff(\ul{u}, p) : \ull{E}(\ul{v})
+    \difd{V} = 0
+    \;, \quad \forall \ul{v} \;,
+
+    \intl{\Omega\suz}{} q\, (J(\ul{u})-1) \difd{V} = 0
+    \;, \quad \forall q \;.
+
+Following formula holds for axial true (Cauchy) stress in the case of uniaxial
```
rc: The following... for the axial true ...

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+    \sigma(\lambda) =
+        2\, \left( C_{(10)} + \frac{C_{(01)}}{\lambda} \right) \,
+        \left( \lambda^2 - \frac{1}{\lambda} \right) \;,
+
+where :math:`\lambda = l/l_0` is the prescribed stretch (:math:`l_0` and
+:math:`l` being the original and deformed specimen length respectively).
+
+The boundary conditions are set so that a state of uniaxial stress is achieved,
+i.e. appropriate components of displacement are fixed on the "Left", "Bottom",
+and "Near" faces and monotonously increasing displacement is prescribed on the
```
rc: a monotonously increasing displacement

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+"Right" face. This prescribed displacement is then used to calculate
+:math:`\lambda` and to convert second Piola-Kirchhoff stress to true (Cauchy)
```
rc: the second Piola-Kirchhoff stress to the true

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,372 @@
...
+
+Different dimensions of the computational domain::
+
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py --dims 2 1 3
+
+Different length of time interval and/or number of time steps::
+
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py -t 0,15,21
+
+Use higher approximation order (this probably needs to reduce time step)::
```
rc: The remark about the time step should be removed (it works well with a proper `i_max`).

rc:
Thanks, I have finally got to run the code - it works well! Just set the max. number of iterations properly, see https://github.com/sfepy/sfepy/pull/443#discussion_r166919697.

Then I have several other comments:

- After some thought, I think it would be better to treat all the multi-value arguments like `--ts`, as that is

  1. consistent with the other examples
  2. allows the compact `--shape=10,5,23' syntax
  
  See the other interactive examples. To parse the values, nothing fancy is needed:
  ```
  options.dims = [float(ii) for ii in options.dims.split(',')]
  ```
- The descriptions of function parameters are sentences and should end with the full-stop ".". Also check the English.
- The function names are not so long to require starting the arguments list on the next line.
- Use the following to reuse the docstring in the help message.
```
from argparse import ArgumentParser, RawDescriptionHelpFormatter
...
    description=__doc__,
                            formatter_class=RawDescriptionHelpFormatter
```
- Some further minor comments were attached to the appropriate lines.

rc:
One more thing: it would be good to rebase your branch on the current master, so that the Travis CI tests pass (some issues related to NumPy 1.14.0 upgrade were just fixed). Let me know if you need help with that.

heczis:
Thanks for the review, I fixed almost all of the points apart from:
- "reuse the docstring in the help message"
  Because the resulting help message is several pages long in my terminal, which seems to me too verbose to be useful.

- "The function names are not so long to require starting the arguments list on the next line."
  PEP8 allows both [1] and I believe it is advantageous to start the argument list on the next line.

- "The remark about the time step should be removed (it works well with a proper `i_max`)."
  It does not work for me, I get a `warp violation` and SfePy gives up. This happens with `i_max=5` (and no other settings to the solver).

I haven't merged the individual commits yet in order to keep track of the changes at the moment, but I expect doing so before merging the PR (in order to keep the history clean).

[1] https://www.python.org/dev/peps/pep-0008/#indentation

rc:
On 02/09/2018 03:05 PM, Jan Heczko wrote:
> Thanks for the review, I fixed almost all of the points apart from:
> - "reuse the docstring in the help message"
>    Because the resulting help message is several pages long in my terminal, which seems to me too verbose to be useful.

Well, the options are printed as the last, so they have the same visibility, 
and there are the usage examples, which is IMHO useful to show. But I do not 
insist on this.

> - "The function names are not so long to require starting the arguments list on the next line."
>    PEP8 allows both [1] and I believe it is advantageous to start the argument list on the next line.

OK, no problem.

> - "The remark about the time step should be removed (it works well with a proper `i_max`)."
>    It does not work for me, I get a `warp violation` and SfePy gives up. This happens with `i_max=5` (and no other settings to the solver).

I see, I misunderstood the remark. In this case I propose "(the ``-t`` option 
to decrease the time step is required for convergence here)" or something like 
that.

> I haven't merged the individual commits yet in order to keep track of the changes at the moment, but I expect doing so before merging the PR (in order to keep the history clean).

This is not needed (the changes are in a branch), but as you wish. Do you mean 
squashing all the commits together, or just some that are related?

> [1] https://www.python.org/dev/peps/pep-0008/#indentation
>

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,392 @@
...
+
+    ax_difference.set_ylabel(r'difference in true stress $\mathrm{[Pa]}$')
+    ax_difference.set_xlabel(r'stretch $\mathrm{[-]}$')
+    ax_difference.grid()
+    plt.show()
+
+def stress_strain(
+        out, problem, _state, order=1, global_stress=None,
+        global_displacement=None, **_
+):
```
rc: Put the "):" to the line above, as in `plot_graphs()`?

heczis:
I meant squashing the related ones, at least to get rid of the unnecessary "Fix ..." messages.

rc:
On 02/10/2018 04:13 PM, Jan Heczko wrote:
> I meant squashing the related ones, at least to get rid of the unnecessary "Fix ..." messages.
> 

OK!

heczis:
I figured out two things:

- I forgot to update the usage examples after changing the multi-value command line arguments. This is fixed now.

- Shuffling and squashing the commits wasn't as easy as I thought, there were a lot of conflicts usually, therefore I only did a few of them.

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,394 @@
...
+In this model, the deformation energy density per unit reference volume is
+given by
+
+.. math::
+    W = C_{(10)} \, \left( \overline I_1 - 3 \right)
+        + C_{(01)} \, \left( \overline I_2 - 3 \right) \;,
+
+where :math:`\overline I_1` and :math:`\overline I_2` are the first
+and second main invariants of the deviatoric part of the right
+Cauchy-Green deformation tensor :math:`\overline{\ull{C}}`. The coefficients
```
rc: Why `\overline` here?

examples/large_deformation/hyperelastic_tl_up_interactive.py
```diff
@@ -0,0 +1,394 @@
+r"""
```
rc: Could you add
```
#!/usr/bin/env python
```
to the first line, and make the script executable (`chmod a+x`), so that it can be executed on Linux without explicitly typing 'python'? All the other interactive examples have that.

rc:
Thanks, almost LGTM! (And yes, squashing the commits (if you do not squash all) can be tedious, and it does not bring much, except maybe a nicer history - you can get the overall changes of a PR without that.)

rc:
LGTM! Merging...

PR #444, opened 2018-02-07
interface to MUMPS linear solver

vlukes:
ToDo: parallel running (?)

### DIFF
--- a/sfepy/solvers/ls.py
+++ b/sfepy/solvers/ls.py
@@ -732,6 +732,66 @@ class PETScKrylovSolver(LinearSolver):
 
         return sol
 
+
+class MUMPSSolver(LinearSolver):
+    """
+    Interface to MUMPS solver.
+
+    """
+    name = 'ls.mumps'
+
+    __metaclass__ = SolverMeta
+
+    _parameters = []
+
+    def __init__(self, conf, **kwargs):
+        try:
+            from mumps import DMumpsContext
+        except ImportError:
+            self.mumps = None
+            msg = 'cannot import MUMPS!'
+            raise ImportError(msg)
+
+        LinearSolver.__init__(self, conf, **kwargs)
+        self.mumps = DMumpsContext()
+        self.mumps_presolved = False
+
+    @standard_call
+    def __call__(self, rhs, x0=None, conf=None, eps_a=None, eps_r=None,
+                 i_max=None, mtx=None, status=None, **kwargs):
+
+        context = self.mumps
+
+        if not self.mumps_presolved:
+            self.presolve(mtx)
+
+        out = rhs.copy()
+        context.set_rhs(out)
+        context.run(job=3)  # Solve
+
+        return out
+
+    def presolve(self, mtx):
+        is_new, mtx_digest = _is_new_matrix(mtx, self.mtx_digest)
+        if is_new:
+            mtx_coo = mtx.tocoo()
+            context = self.mumps
+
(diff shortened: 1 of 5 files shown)


rc:
Great! That was on my TODO for some time already :) I will check it soon, For now I have just some suggestions/questions:

- What about calling it just `ls.mumps`? (like `ls.petsc', in contrast with 'ls.pyamg', which is because the solver _is_ called PyAMG...)
- The `silent` option is not needed - all solvers have `verbose` option, that is by default False.
- Is it possible to prefactorize a matrix and then reuse it with different right-hand sides?
- How difficult it is to add the parallel capabilities? We can assemble in parallel using PETSc, so I wonder if we could reuse that here.

BTW. I am just fixing the Travis CI failures related to numpy 0.14.0 - I will let you know once the fixes are merged, and then, please, rebase this branch.

vlukes:
On 02/07/2018 03:18 PM, Robert Cimrman wrote:
> Great! That was on my TODO for some time already :) I will check it 
> soon, For now I have just some suggestions/questions:
> 
>   * What about calling it just |ls.mumps|? (like `ls.petsc', in contrast
>     with 'ls.pyamg', which is because the solver /is/ called PyAMG...)
Yes, `ls.mumpy` will be better.

>   * The |silent| option is not needed - all solvers have |verbose|
>     option, that is by default False.
I will use verbose instead of silent.

>   * Is it possible to prefactorize a matrix and then reuse it with
>     different right-hand sides?
This is already working. See, `context.run(job=X)`, where X is 
1=analyze, 2=factorize, 3=solve.

>   * How difficult it is to add the parallel capabilities? We can
>     assemble in parallel using PETSc, so I wonder if we could reuse that
>     here.

That the question. As I know, parallel MUMPS solver needs to be run in 
the MPI context and I don't know (for now) how to do that in Python. 
Something like "mpirun XY_function()".

> BTW. I am just fixing the Travis CI failures related to numpy 0.14.0 - I 
> will let you know once the fixes are merged, and then, please, rebase 
> this branch.
> 
OK.

rc:
Actually, for the parallel solves, we could just use MUMPS through PETSc (via `ls.petsc`) - it is interfaced there (http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MATSOLVERMUMPS.html). Of course, PETSc would have to be built with MUMPS support.

rc:
Ad pre-factorization:  could you move the pre-factorization code into a `.presolve()` method (see `ScipyDirect`), so that it could be used from `Problem.try_presolve()`?

rc:
The NumPy 1.14.0 related fixes are merged...

vlukes:
The suggestions are incorporated. Parallel run is still an open task.

sfepy/solvers/ls.py
```diff
@@ -732,6 +732,66 @@ def __call__(self, rhs, x0=None, conf=None, eps_a=None, eps_r=None,
...
+
+    def __init__(self, conf, **kwargs):
+        try:
+            from mumps import DMumpsContext
+        except ImportError:
+            self.mumps = None
+            msg = 'cannot import MUMPS!'
+            raise ImportError(msg)
+
+        LinearSolver.__init__(self, conf, mg=None, **kwargs)
```
rc: What is the `mg` argument for? It seems unused.

vlukes:
Capy/Paste from pyamg solver, I will remove it.

Issue #447, opened 2018-02-16
generate tables of availables solvers


rc:
Similarly to the automatically generated tables of terms, tables of the various solvers could be generated to be included in the docs.

See `script/gen_term_table.py` and `doc/conf.py`.

BubuLK:
Working on this issue (no duplicated work, please ;), PR will follow shortly, be patient...

rc:
Great, thanks!

rc:
BTW. I have just merged a branch adding new solvers - you may want to rebase...

PR #451, opened 2018-03-19
Generate table of available solvers

BubuLK:
First attempt to add auto-generated table (not "table" actually...) of available solvers (#447).
This is still work in progress, any suggestions are welcome...


doc/users_guide.rst
```diff
@@ -1358,13 +1358,13 @@ The following code demonstrates the use of the postprocessing filters::
...
------------------
+Solvers
+-------
 
-This Section describes the linear, nonlinear, eigenvalue, optimization and
-time-stepping solvers available in SfePy. There are many internal and external
-solvers in the :ref:`sfepy_solvers` that can be called using a
-uniform interface.
+This section describes the time-stepping, linear, nonlinear, eigenvalue and
+optimization and solvers available in *SfePy*. There are many internal and
```
rc: optimization and -> optimization

rc:
Great, thanks!

I suggest to shorten the description of each solver to the first paragraphs of the solver class docstrings. Some of the docstring will need some care, and could be updated with the information currently available in the hand-written list of solvers. I can help you with that. Then the text generated from docstrings could just replace the hand-written one, so that there is a single source of information and no need to keep it in sync manually. One can include a rst file, see http://docutils.sourceforge.net/docs/ref/rst/directives.html#include, so maybe we could generate a rst for each kind of solvers and include it in its section in the users' guide.

Concerning the generating script, maybe it could reuse ``solver_table`` from ``sfepy.solvers.__init__.py`` directly?

rc:
Thanks for the updates, this is what I had in mind :) Only some minor issues remain:

- The section ordering should be time-stepping, nonlinear, linear - that is the sequence how the outer solvers call inner solvers when a problem is solved. The first sentence in Solvers section could reflect that too.
- Sort alphabetically the entries in each list.
- Instead of "the first three solvers" in the linear solvers section, either remove the sentence, or write explicitly: ls.scipy_direct, ls.scipy_iterative, ls.petsc.
- The eigenvalue problem solvers and optimization solvers should have their own sections (in this order).
- As you said, the first paragraphs of some solvers need updating (I will help with this).

BubuLK:
Updated according to comments.
TBD:
  - some descriptions missing (new solver-type sections)
  - docstrings should be updated to follow PEP 257 guidelines...

doc/users_guide.rst
```diff
@@ -1400,43 +1399,52 @@ always has zero boundary conditions.
...
+The following solvers are available:
+
+.. include:: solver_table.rst
+    :start-after:   .. <Linear Solvers>
+    :end-before:    .. </Linear Solvers>
+
+See :mod:`sfepy.solvers.ls` for all available *Linear* solvers and their
+options.
+
+Eigen Value Solvers
```
rc: Eigenvalue Problem Solvers

rc:
Thanks, I will try to fill in the missing texts.

We are using the NumPy docstring style, and we like it ;)

rc:
@BubuLK I have fixed the docstrings and made some small changes in https://github.com/rc/sfepy/tree/gen-solvers-table, so fetch those changes into your branch and push it here.

BubuLK:
Done. (Probably not in the right way from git POV, sorry ;()

rc:
Great, thanks - LGTM!

Just one last thing got on my mind: could you make the solver names in the lists link to the corresponding solver class documentation? The following syntax
```
 :class:`solver_name <full.path.to.Class>`
```
could be used for that.

BubuLK:
Done (if I understood well).

rc:
Yes, thanks!

PR #462, opened 2018-08-23
fix Problem.setup_output(): allow dots in domain name

vlukes:


rc:
I guess this does not conflict with the fact, that `self.ofn_trunk` should be the mesh file name without a suffix. This is for manually named domains, right?

LGTM, just rebase onto the current master.

vlukes:
Yes, I need it for a manually named domain. Rebased.

PR #463, opened 2018-08-24
fix extract_surface.py: remove useless semicolon

vlukes:

PR #464, opened 2018-08-24
new_extract_edges.py: extract outline edges of a given mesh

vlukes:


script/extract_edges.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
+"""
+Extrtact mesh edges and save them into
```
rc: Extrtact mesh edges -> Extract surface edges of a mesh

rc:
Thanks! I have some comments:

- Maybe you could use CMesh.get_conn_as_graph()?
- What does it exactly do in 3D - what are the normals for?
- Could you rebase this PR as well as #462 and #463 on the current master?

rc:
I have finally tried the script. When used as
```
./script/extract_edges.py cylinder.mesh
```
it creates `cylinder_edges.vtk`, which does not contain the edges on the two sides:
edges
Is this correct?

vlukes:
Yes, your result is correct. Each of the edge has two incident facets, if the difference of the normal vectors is grater than `eps`, the edge is labeled as the outline edge. If you run
 ```
./script/extract_edges.py --eps 1e-12 cylinder.mesh
```
you get the result without edges at the bottom/top part. I have increased the default value of `eps` to 1e-12.

I have implemented `merge_lines()` function which merges all edges laying in a line into the single edge.

File: script/extract_edges.py [hunk 2]
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env python
+"""
+Extrtact outline edges of a given mesh and save them into
+'<original path>/edge_<original mesh file name>.vtk'
+or into a user defined output file.
+The outline edge is a edge for which norm(nvec1 - nvec2) < eps,
+where nvec1 and nvec2 are the normal vectors of the incident facets.
+"""
```
rc: Corrections: Extrtact -> Extract, a edge -> an edge

vlukes:
Fixed

script/extract_edges.py
```diff
@@ -0,0 +1,214 @@
...
+    Extract outline edges of a given mesh.
+    The outline edge is an edge for which norm(nvec_1 - nvec_2) < eps,
+    where nvec_1 and nvec_2 are the normal vectors of the incident facets.
+
+    Parameters
+    ----------
+    mesh : Mesh
+        The 3D or 2D mesh.
+    eps : float
+        The tolerace parameter of the outline edge searching algorithm.
```
rc: tolerance

script/extract_edges.py
```diff
@@ -0,0 +1,214 @@
...
+
+        mesh_out = ed_coors, ngroups, [ed_conns], [mat_ids], ['3_2']
+
+        return mesh_out
+
+    else:
+        raise ValueError('no outline edges found (eps=%e)!' % eps)
+
+helps = {
+    'eps': 'tolerace parameter of the edge search algorithm (default: 1e-12)',
```
rc: tolerance

rc:
Thanks, also do not forget to add the script to aux_scripts in setup.py, so that it gets into the release tarballs.

PR #467, opened 2018-10-10
update linear_homogenization.py: set option 'is_linear' to True for s…

vlukes:
…peed-up


rc:
BTW. I am aware of the fact that this example is not covered by the tests, so I checked it manually. Note that MUMPS is not a required dependency, so using it as the solver here is a bit on the edge - there should be a fallback option.

vlukes:
It was not the aim to change the solver in this example, I just forget to remove the line and uncomment the line with `scipy_direct`.

vlukes:
...  just forgot to ...

rc:
OK, now it's better :)

PR #468, opened 2018-10-10
fix mumps solver and update for mumps library version 5.1.2

vlukes:


rc:
Thanks! Could you rebase both this and #467 onto the current master?

Also, what is your compatibility policy with MUMPS versions? Does the code work also with previous ones?

vlukes:
- Both the requests are now rebased onto the current master.

- The base C structure has changed since the major version 5 of MUMPS, so there are defined two different structures (one for 4.10.x and another one for 5.1.x) and which are switched according to the installed version of the library.  I suppose (I really don't know) that the interface (C structure) is stable in lines 4 and 5. So, if the version is < 5.0 then mumps_struc_c_4.10.1 is used, otherwise mumps_struc_c_5.1.2.

PR #472, opened 2018-10-15
Fix splinebox

vlukes:


rc:
Thanks!

Could you remove the 'update coeffficients.py' commit (it is in #471, right?)
How does the new code behave with NumPy < 1.14.0?

rc:
Also why does tests/test_homogenization_perfusion.py fail in Python 3?

vlukes:
'update coeffficients.py' commit is removed.

rc:
The problem disappeared, let's merge this :)

rc:
Yes a fix is needed, the current version of calling `lstsq()` does not work with numpy 1.13.3.

Issue #411, opened 2017-07-19
use asv for benchmarks


rc:
See https://asv.readthedocs.io/en/latest/index.html.

PR #475, opened 2018-10-26
update mumps: use a symmetric solver if the matrix is symmetric

vlukes:
See the solution times for UMPACK, MUMPS - unsymmetric solver  and MUMPS - symmetric solver.

sfepy: matrix shape: (84861, 84861)
sfepy: matrix structural nonzeros: 6637127 (9.22e-04% fill)

umfpack_vs_mumps

What about using MUMPS as the default linear solver?

### DIFF
--- a/sfepy/solvers/ls.py
+++ b/sfepy/solvers/ls.py
@@ -767,10 +767,14 @@ class MUMPSSolver(LinearSolver):
         return out
 
     def presolve(self, mtx):
+        if not isinstance(mtx, sps.coo_matrix):
+            mtx = mtx.tocoo()
         if self.mumps_ls is None:
             system = 'complex' if mtx.dtype.name.startswith('complex')\
                 else 'real'
-            self.mumps_ls = self.mumps.MumpsSolver(system=system)
+            is_sym = self.mumps.coo_is_symmetric(mtx)
+            self.mumps_ls = self.mumps.MumpsSolver(system=system,
+                                                   is_sym=is_sym)
 
         is_new, mtx_digest = _is_new_matrix(mtx, self.mtx_digest)
         if is_new:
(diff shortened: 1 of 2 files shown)


rc:
Thanks for the PR. Nice speed-up with MUMPS in comparison with UMFPACK!

I would expect more speed-up with the symmetric version - does the conversion to COO format pose a significant overhead? Maybe more in terms of memory then speed?

As for making MUMPS the default solver - I am not against, but several points needs to be addressed:

- installation issues: is it in conda/other package systems, is it easy to install on windows or os x?
- make it a required run-time dependency, or just optional?
- if UMFPACK is not available, the `ls.scipy_direct` solver falls back to the super-LU built in scipy. How to provide a fallback if MUMPS is optional?
- In any case, the docs would need to be extensively updated.
- Something else?

A good thing is that no external package is required to install the Python wrappers :)

vlukes:
I would also expected a higher speed-up with the symmetric solver. The COO conversion is very fast and the overhead is negligible. Memory usage by MUMPS with the unsymmetric matrix is 1228MB while the with the symmetric matrix only 759MB. So the memory saving is significant.

MUMPS as default solver - I will create a new issue and we can discuss it there.

rc:
> I would also expected a higher speed-up with the symmetric solver. The COO conversion is very fast and the overhead is negligible. Memory usage by MUMPS with the unsymmetric matrix is 1228MB while the with the symmetric matrix only 759MB. So the memory saving is significant.

But note that the COO matrix is in addition to the original CSR.

It might be interesting to run your benchmark with memory_profiler. Also, adding some benchmarks to sfepy seems like a good idea to me, see #411. Some steps were already  made towards it, but we got distracted by other topics - @BubuLK, could you, please, make a PR with your current code, so that it is not lost?
 
> MUMPS as default solver - I will create a new issue and we can discuss it there.

 OK

PR #482, opened 2018-11-15
Update linear solvers, new solvers with auto fallback

vlukes:
**New:**
* new `ScipySuperLU`, `ScipyUmfpack` linear solvers - shortcuts to  `ScipyDirect(method='superlu')` and `ScipyDirect(method='umfpack')`
* `fallback` option for linear solvers
* `AutoFallbackSolver` class - fallback chain given by ` _ls_solvers` list
* `AutoDirect` and `AutoIterative` (which solvers/methods should be used?) solvers

**Updated:**
* `MUMPSSolver`,  `MUMPSParallelSolver` - consider matrix symmetry -> faster factorization and lower memory demands
* `MUMPSParallelSolver` - accept verbose option
* examples - using `AutoDirect` and `AutoIterative`
* documentation related to the above

### DIFF
--- a/sfepy/discrete/problem.py
+++ b/sfepy/discrete/problem.py
@@ -26,7 +26,7 @@ from sfepy.discrete.evaluate import create_evaluable, eval_equations
 from sfepy.solvers.ts import TimeStepper
 from sfepy.discrete.evaluate import Evaluator
 from sfepy.solvers import Solver, NonlinearSolver
-from sfepy.solvers.solvers import ls_fallback
+from sfepy.solvers.solvers import use_first_available
 from sfepy.solvers.ts_solvers import StationarySolver
 import six
 from six.moves import range
@@ -1069,7 +1069,7 @@ class Problem(Struct):
                     break
 
             if len(fb_list) > 1:
-                ls = ls_fallback(fb_list, context=self)
+                ls = use_first_available(fb_list, context=self)
             else:
                 ls = Solver.any_from_conf(ls_conf, context=self)
 
--- a/sfepy/discrete/problem.py
+++ b/sfepy/discrete/problem.py
@@ -26,6 +26,7 @@ from sfepy.discrete.evaluate import create_evaluable, eval_equations
 from sfepy.solvers.ts import TimeStepper
 from sfepy.discrete.evaluate import Evaluator
 from sfepy.solvers import Solver, NonlinearSolver
+from sfepy.solvers.solvers import ls_fallback
 from sfepy.solvers.ts_solvers import StationarySolver
 import six
 from six.moves import range
@@ -985,18 +986,27 @@ class Problem(Struct):
     def set_conf_solvers(self, conf_solvers=None, options=None):
         """
         Choose which solvers should be used. If solvers are not set in
-        `options`, use first suitable in `conf_solvers`.
+        `options`, use the ones named `ls`, `nls` or `ts`. If such solvers
+        do not exist, us the firsts suitable in `conf_solvers`.
         """
         conf_solvers = get_default(conf_solvers, self.conf.solvers)
         self.solver_confs = {}
+
         for key, val in six.iteritems(conf_solvers):
             self.solver_confs[val.name] = val
 
         def _find_suitable(prefix):
+            cands = []
             for key, val in six.iteritems(self.solver_confs):
                 if val.kind.find(prefix) == 0:
-                    return val
-            return None
+                    if val.name == prefix[:-1]:
+                        return val
+                    else:
+                        cands.append(val)
+            if len(cands) > 0:
+                return cands[0]
+            else:
+                return None
 
         def _get_solver_conf(kind):
             try:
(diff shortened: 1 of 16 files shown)


rc:
Thanks, it looks good!

I have some corrections to make (typos, etc.) that I would rather do as separate commits instead of just comments (it's more efficient to do a thing directly than describing it).  I can put a branch to my github account, or I could push directly into your branch (if you allow it) - what do you think/prefer?

Some things to consider:

- Instead of ls_fallback(), what about allowing fallback lists in Solver.any_from_conf()?
- Then ls_fallback() could be used internally in Solver.any_from_conf(). But this function is not tied to linear solvers only, so it could be called just fallback()? Or create_available()? Or?

sfepy/discrete/problem.py
```diff
@@ -985,18 +986,27 @@ def get_ebc_indices(self):
...
-                    return val
-            return None
+                    if val.name == prefix[:-1]:
+                        return val
+                    else:
+                        cands.append(val)
+            if len(cands) > 0:
+                return cands[0]
+            else:
+                return None
```
rc: What is the reason of this change? I mean, why creating a list, when the first item is always used.

rc:
Besides the question above, I have some more updates in https://github.com/rc/sfepy/tree/update_linear_solvers - could you check that? Thanks!

sfepy/discrete/problem.py
vlukes: This change ensures to select either `ls`/`nls`/`ts`, as required by `_get_solver_conf()`, if it exists, or the first linear/nonlinear/time solver in the created list.

Examples:
 required 'ls', defined solvers ['ls3', 'ls2', 'ls']  -> 'ls'
 required 'ls', defined solvers ['ls3', 'ls2', 'ls1']  -> 'ls3'

vlukes:
The updates are OK. Thanks.

sfepy/discrete/problem.py
rc: OK, so you probably do not need to append to the list - just replacing the first item would be ok (anything mutable would work), but nevermind...

rc:
BTW. include my commits in you branch... (before the rebase?) Thanks!

vlukes:
Commits included.

rc:
Thanks! I have one more fix in my branch for missing petsc4py - can you include it here?

rc:
It looks that MUMPS wrappers have some problems on windows:
```
 File "run_tests.py", line 226, in main
    run_tests(stats, dirpath, filenames)
  File "run_tests.py", line 144, in run_tests
    stats[0] + 1)
  File "run_tests.py", line 108, in run_test
    ifile=ifile)
  File "C:\projects\sfepy\sfepy\base\testing.py", line 67, in run
    ret = test_method()
  File "C:\projects\sfepy\tests\tests_basic.py", line 85, in test_input
    post_process_hook_final=self.post_process_hook_final)
  File "C:\projects\sfepy\sfepy\applications\pde_solver_app.py", line 63, in solve_pde
    return app(status=status)
  File "C:\projects\sfepy\sfepy\applications\application.py", line 29, in call_basic
    return self.call(**kwargs)
  File "C:\projects\sfepy\sfepy\applications\pde_solver_app.py", line 225, in call
    post_process_hook_final=self.post_process_hook_final)
  File "C:\projects\sfepy\sfepy\discrete\problem.py", line 1366, in solve
    self.init_solvers(status=status)
  File "C:\projects\sfepy\sfepy\discrete\problem.py", line 1075, in init_solvers
    ls = Solver.any_from_conf(ls_conf, context=self)
  File "C:\projects\sfepy\sfepy\solvers\__init__.py", line 29, in any_from_conf
    return solver_table[conf.kind](conf, **kwargs)
  File "C:\projects\sfepy\sfepy\solvers\auto_fallback.py", line 24, in __new__
    return use_first_available(ls_solvers)
  File "C:\projects\sfepy\sfepy\solvers\solvers.py", line 90, in use_first_available
    out = Solver.any_from_conf(_conf, context=context)
  File "C:\projects\sfepy\sfepy\solvers\__init__.py", line 29, in any_from_conf
    return solver_table[conf.kind](conf, **kwargs)
  File "C:\projects\sfepy\sfepy\solvers\ls.py", line 787, in __init__
    mumps.load_mumps_libraries()  # try to load MUMPS libraries
  File "C:\projects\sfepy\sfepy\solvers\ls_mumps.py", line 43, in load_mumps_libraries
    mumps_libs['dmumps'] = load_library('dmumps').dmumps_c
  File "C:\projects\sfepy\sfepy\solvers\ls_mumps.py", line 37, in load_library
    lib = ctypes.cdll.LoadLibrary(lib_fname)
  File "C:\Python36_64\envs\sfepy-test\lib\ctypes\__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "C:\Python36_64\envs\sfepy-test\lib\ctypes\__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
TypeError: LoadLibrary() argument 1 must be str, not None
```

I will catch the TypeError for the moment to see if there are further problems.

rc:
The last appveyor failure with Python 2.7 (3.6 works!) is related to using numpy.block() that was added in 1.13.0 - the tests use a prior version. So let us merge this! (And raise minimal sfepy requirements.)

PR #484, opened 2018-11-16
fix ls.mumps for symmetric complex valued matrices

vlukes:


rc:
Thanks! Later it would be nice to have tests for complex problems (maybe in test_linear_solvers.py? It needs a facelift anyway).

PR #454, opened 2018-05-21
Allow empty dict in dict_from_string

lokik:

PR #485, opened 2018-11-23
Parse conf

lokik:
Fixes for parse conf - allow empty dict, empty lists etc....

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -141,15 +141,11 @@ class Test( TestCommon ):
 
     def test_parse_conf(self):
         from sfepy.base.parse_conf import list_dict
-        from sfepy.base.conf import dict_from_string
+        from sfepy.base.conf import dict_from_string as parse_dict
         ld = list_dict()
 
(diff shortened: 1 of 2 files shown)


rc:
Thanks! What is the relation of this to #454?

tests/test_base.py
```diff
@@ -141,16 +141,36 @@ def test_resolve_deps(self):
...
     def test_parse_conf(self):
         from sfepy.base.parse_conf import list_dict
+        from sfepy.base.conf import dict_from_string
         ld = list_dict()
 
         def parse(x):
             out = ld.parseString(x, True)[0]
             return out
 
+        def parse_dict(x):
```
rc: Just use:
```
from sfepy.base.conf import dict_from_string as parse_dict
```
and remove this function.

rc:
So it seems to address the same thing as  #454 is a more thorough way - then #454 can be closed, right?

rc:
Thanks, LGTM.

PR #487, opened 2018-11-23
init_linear_solver

lokik:
Allows user to specify how to create linear solver in Problem (e.g. define fallback if solver requested from conf is not available).


rc:
Are you aware of #482? :) It looks like it might solve the same thing.

PR #488, opened 2018-11-26
do not execute code in ls_mumps_parallel.py on import

rc:
@vlukes what do you think of this change? Running the code on import causes an error when generating the sphinx docs.


vlukes:
Yes, that seems reasonable.

rc:
Thanks, I will merge it then.

PR #490, opened 2018-12-04
Update example in the docstring of Problem.create_evaluable

heczis:
The example uses a term that already changed its name twice over the years  (`dq_state_in_volume_qp`).


rc:
Thanks!

PR #492, opened 2018-12-11
fix presolve() in  mumps solver

vlukes:


sfepy/solvers/ls.py
```diff
@@ -783,7 +786,8 @@ def presolve(self, mtx):
 
             self.mumps_ls.set_mtx_centralized(mtx)
             self.mumps_ls(4)  # analyze + factorize
-            self.mumps_presolved = True
+            if presolve_flag:
+                self.mumps_presolved = True
```
rc: It looks like the matrix is pre-factorized in any case, so the presolve option only means, that the solution can be re-used?

Issue #494, opened 2019-01-24
Multiprocessing implementation on Windows is broken


BubuLK:
Multiprocessing implementation (proc version) in `sfepy/homogenization/engine.py` is broken on Windows platform.
Apply to both Python 2.7 and 3.6 versions.

There could be different error messages, but all leads to some kind of "pickling error" during **spawn** worker process (in contrast with **fork** on POSIX systems).

rc:
See also the fixes in #495.

Issue #476, opened 2018-10-27
failing tests in latest build


awa5114:
I have tried running the tests on what I think is the latest version of sfepy but got 5 failures out of 174 tests. Is this being addressed? Below is the failing test list:

 - test_homogenization_perfusion.py
 - test_input_linear_elastic_mM.py
 - test_input_navier_stokes2d_iga.py
 - test_hdf5_meshio
 - test_projection_iga_fem

Also how can we determine what revision broke these tests?

vlukes:
Hi,

 >   * test_input_navier_stokes2d_iga.py
 >   * test_hdf5_meshio
 >   * test_projection_iga_fem

failed due to the missing igakit module and

 >   * test_homogenization_perfusion.py
 >   * test_input_linear_elastic_mM.py

probably failed because of wrong MPI configuration. Try to add

'multiprocessing': False

into 'options' in the problem description file. It should fix the issue. 
If not, run

./run_tests.py --debug tests/test_homogenization_perfusion.py

and send me the output.


Regards,
Vladimir

awa5114:
In `perfusion_micro.py` I added `'multiprocessing':False` in the `options` and got the following error

```
(sfepy_env) D:\sfepy\sfepy>run_tests.py tests/test_homogenization_perfusion.py
Traceback (most recent call last):
  File "D:\sfepy\sfepy\run_tests.py", line 32, in <module>
    from sfepy.base.conf import ProblemConf, get_standard_keywords
  File "D:\sfepy\sfepy\sfepy\base\conf.py", line 14, in <module>
    from sfepy.base.base import (Struct, IndexedStruct, dict_to_struct,
  File "D:\sfepy\sfepy\sfepy\base\base.py", line 11, in <module>
    import scipy.sparse as sp
ImportError: No module named scipy.sparse
```

I'm also a bit confused that there are missing modules. I went through the following steps to run a development version of sfepy:

- create new `sfepy_env` environment
- install sfepy with `conda install -c conda-forge sfepy` to get necessary dependencies
- remove sfepy with `conda remove sfepy`
- fork & clone a copy of the repo to folder of my choosing
- run an in-place build `python setup.py build_ext --inplace`
- run tests through `sfepy_env`

I would have thought dependencies such as  your `igakit` module and `scipy` would have been taken care of in step 2 above. Why are the tests still complaining about them?

vlukes:
The `igakit` module is optional. If you don't need the isogeometric 
[...]

PR #495, opened 2019-01-24
Tests suite "fix"

BubuLK:
Fixed some long-lasting problems (#491 or #476 ) in tests suite:

- added tests for missing/non-working IGAkit (all platforms)
- disabled broken multiprocessing on windows platform (#494 )
- some additional fixes on windows platform

TBD  (MacOS platform):
- test_refine_hanging.py is "flapping" (OK/seg-fault). Not easy to catch/debug... ;(


BubuLK:
> As for the test_refine_hanging.py problem, I am not sure how to debug that, besides the usual advice like adding print statements everywhere, examining the core (compiling with debug symbols), etc.
> 
> Are you on the track of the problem, or should I merge this PR, and the remaining issues will be handled later (if possible)?

Probably merging this PR is better... (If this problem will remain, I'll investigate later on...

PR #500, opened 2019-02-05
Surface integrate material term

rc:
Just added `ev_surface_integrate_mat`, as `ev_integrate_mat` has been volume-only. I have also added an alias `ev_volume_integrate_mat` to `ev_integrate_mat`.

I wonder if we should simply replace `ev_integrate_mat` with `ev_volume_integrate_mat`, what do you think, @vlukes?


vlukes:
I prefer to replace `ev_integrate_mat` with `ev_volume_integrate_mat`.

rc:
Same here, I will replace the name, thanks for the reply.

PR #506, opened 2019-04-01
Homogenization - Fading Memory infrastructure 

vlukes:
Revision of FM coefficients/correctors:

- New `TSCoef` coefficient that allows to replace `CoefFM*` by `TSCoef(Coef*)`, e.g. `CoefFMSymSym` --> `TSCoef(CoefSymSym)`, `CoefFMOne` --> `TSCoef(CoefOne)`
- EVP correctors return `CorrSolution` structure as the other ones
- Fixed `get_time_solver()`

Fixed the following issues related to the linear solvers:
- `presolve()` method renamed to 'presolve_fun()', name conflict with attribute `presolve`
- fixed `presolve()` in `MUMPSSolver`

### DIFF
--- a/sfepy/homogenization/coefs_base.py
+++ b/sfepy/homogenization/coefs_base.py
@@ -910,7 +910,8 @@ class TSTimes(MiniAppBase):
     """Coefficient-like class, returns times of the time stepper."""
     def __call__(self, volume=None, problem=None, data=None):
         problem = get_default(problem, self.problem)
-        return problem.get_time_solver().ts.times
+        problem.init_solvers()
+        return problem.get_timestepper().times
 
 class VolumeFractions(MiniAppBase):
     """Coefficient-like class, returns volume fractions of given regions within
--- a/sfepy/homogenization/coefs_base.py
+++ b/sfepy/homogenization/coefs_base.py
@@ -181,7 +181,7 @@ class CorrMiniApp(MiniAppBase):
                         var = variables[var_name]
                         shape = (var.n_dof // var.n_components,
                                  var.n_components)
-                        out[skey] = Struct(name = 'dump', mode = 'nodes',
+                        out[skey] = Struct(name = 'dump', mode = 'vertex',
                                            data = dof_vector,
                                            dofs = var.dofs,
                                            shape = shape,
@@ -713,6 +713,8 @@ class TCorrectorsViaPressureEVP(CorrMiniApp):
         iee_e_qg = 0.0
         format = '====== time %%e (step %%%dd of %%%dd) ====='\
                  % ((ts.n_digit,) * 2)
+        vu, vp = self.dump_variables
+        state = {k: [] for k in [vu, vp, 'd' + vp]}
         for step, time in ts:
             output(format % (time, step + 1, ts.n_step))
 
@@ -737,7 +739,13 @@ class TCorrectorsViaPressureEVP(CorrMiniApp):
 ##             print aaa
 ##             print bbb
 
+            state[vu].append(vec_u)
+            state[vp].append(vec_p)
+            state['d' + vp].append(vec_dp)
+
             self.save(dump_name, save_name, vec_u, vec_p, vec_dp, ts, problem)
+
+        return state
 
     def save(self, dump_name, save_name, vec_u, vec_p, vec_dp, ts, problem):
         """
@@ -745,12 +753,12 @@ class TCorrectorsViaPressureEVP(CorrMiniApp):
         2. saves correctors transformed to output for visualization
         """
         vu, vp = self.dump_variables
-        out = {vu : Struct(name='dump', mode='nodes', data=vec_u,
-                           dofs=None, var_name=vu),
-               vp : Struct(name='dump', mode='nodes', data=vec_p,
-                           dofs=None, var_name=vp),
-               'd'+vp : Struct(name='dump', mode='nodes', data=vec_dp,
-                               dofs=None, var_name=vp)}
+        out = {vu : Struct(name='dump', mode='vertex',
+                           var_name=vu, data=vec_u[:, nm.newaxis]),
+               vp : Struct(name='dump', mode='vertex',
+                           var_name=vp, data=vec_p[:, nm.newaxis]),
+               'd'+vp : Struct(name='dump', mode='vertex',
+                               var_name=vp, data=vec_dp[:, nm.newaxis])}
 
         problem.save_state(dump_name, out=out, file_per_var=False, ts=ts)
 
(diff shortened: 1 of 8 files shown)


sfepy/homogenization/coefs_base.py
```diff
@@ -737,20 +739,26 @@ def compute_correctors(self, evp, sign, state0, ts, dump_name, save_name,
...
         2. saves correctors transformed to output for visualization
         """
         vu, vp = self.dump_variables
-        out = {vu : Struct(name='dump', mode='nodes', data=vec_u,
-                           dofs=None, var_name=vu),
-               vp : Struct(name='dump', mode='nodes', data=vec_p,
-                           dofs=None, var_name=vp),
-               'd'+vp : Struct(name='dump', mode='nodes', data=vec_dp,
-                               dofs=None, var_name=vp)}
+        out = {vu : Struct(name='dump', mode='nodes', dofs=None,
```
rc: Maybe also remove the dofs arg (not used for anything), and fix `mode='nodes'` to `mode='vertex'`.

rc:
Thanks for the clean-up of the time-dependent correctors!

After looking at the changes in more detail, I do not see the conflict between the presolve option (a part of configuration dict, only in some solvers) and the presolve() function. Moreover, presolve() as a function is also used elsewhere (e.g. in PressureEigenvalueProblem), so I suggest reverting this. (BTW. MUMPSSolver uses presolve_flag as its attribute, but I would leave the status quo)

Using the TSCoef name for a function and not a for class got me confused - could you call it, e.g. create_tscoef()? Also, an example using it would be nice, to see, how it replaces the FM coefficients. I assume you define a class with a single __call__() method, that is passed to TSCoef(), right? In that case it might be simpler to just add a 'call' function attribute (actually solve_step() might be a better name) to the coefficient configuration, no need to parametrize TSCoefClass by inheritance. What do you think?

vlukes:
In general, it is not a good idea to have the same name for two things, an attribute/parameter and a method name. In this case,  `presolve` options can propagate to the namespace as an attribute in  `Solver.__init__()` and it can be in conflict with `presolve()` function. In my configuration, this conflict has already appeared. It is the reason for this commit.

I consider `TSCoef` as  a "class decorator". It returns the class which inherits all the members from the original one but it modifies the `call()` method to accept data in time levels. I hope that this implementation is clear and as simple as possible. Yes, the name is the question. The `TSCoef` name, i.e. the class naming convention, is used now because it returns the class. But, maybe, `create_TSCoef` would be better, it reflects the situation more accurately.

sfepy/homogenization/coefs_base.py
vlukes: OK

rc:
> In general, it is not a good idea to have the same name for two things, an attribute/parameter and a method name. In this case, `presolve` options can propagate to the namespace as an attribute in `Solver.__init__()` and it can be in conflict with `presolve()` function. In my configuration, this conflict has already appeared. It is the reason for this commit.

How did the propagation happen?

Overall I agree, that separate names for separate things are better. In this case, I would preserve the method name (there are *solve() functions/methods elsewhere in sfepy), and rename the option. Maybe use_presolve?

> I consider `TSCoef` as a "class decorator". It returns the class which inherits all the members from the original one but it modifies the `call()` method to accept data in time levels. I hope that this implementation is clear and as simple as possible. Yes, the name is the question. The `TSCoef` name, i.e. the class naming convention, is used now because it returns the class. But, maybe, `create_TSCoef` would be better, it reflects the situation more accurately.

You mean that you can pass the other Coef* in, so that those are usable for time-dependent problems? Now you can see, why an example would be nice. :) OK I get the idea, it adapts the classes for time-stepping: I would call it adapt_for_ts().

vlukes:
> > In general, it is not a good idea to have the same name for two things, an attribute/parameter and a method name. In this case, `presolve` options can propagate to the namespace as an attribute in `Solver.__init__()` and it can be in conflict with `presolve()` function. In my configuration, this conflict has already appeared. It is the reason for this commit.
> 
> How did the propagation happen?

`PressureEigenvalueProblem.presolve()` calls  `Solver.any_from_conf(presolve=True)` and it leads to the name conflict. Maybe, it is incorrectly initialized, but this was the reason for my commit. 

> 
> Overall I agree, that separate names for separate things are better. In this case, I would preserve the method name (there are *solve() functions/methods elsewhere in sfepy), and rename the option. Maybe `use_presolve`?

OK, I will rename the option.

> 
> > I consider `TSCoef` as a "class decorator". It returns the class which inherits all the members from the original one but it modifies the `call()` method to accept data in time levels. I hope that this implementation is clear and as simple as possible. Yes, the name is the question. The `TSCoef` name, i.e. the class naming convention, is used now because it returns the class. But, maybe, `create_TSCoef` would be better, it reflects the situation more accurately.
> 
> You mean that you can pass the other `Coef*` in, so that those are usable for time-dependent problems? Now you can see, why an example would be nice. :) OK I get the idea, it adapts the classes for time-stepping: I would call it `adapt_for_ts()`, or something like that.

Yes, you can pass "any" other `Coef*` into it. What about `create_ts_coef`?  `adapt_for_ts()`is not bad, but ...

rc:
> > > In general, it is not a good idea to have the same name for two things, an attribute/parameter and a method name. In this case, `presolve` options can propagate to the namespace as an attribute in `Solver.__init__()` and it can be in conflict with `presolve()` function. In my configuration, this conflict has already appeared. It is the reason for this commit.
> > 
> > 
> > How did the propagation happen?
> 
> `PressureEigenvalueProblem.presolve()` calls `Solver.any_from_conf(presolve=True)` and it leads to the name conflict. Maybe, it is incorrectly initialized, but this was the reason for my commit.

Yes, the first argument should be a conf, as in:
```
    ls = Solver.any_from_conf(Struct(**{'name' : 'ls', 'kind' : 'ls.auto_direct',}))
```
 
> > Overall I agree, that separate names for separate things are better. In this case, I would preserve the method name (there are *solve() functions/methods elsewhere in sfepy), and rename the option. Maybe `use_presolve`?
> 
> OK, I will rename the option.
> 
> > > I consider `TSCoef` as a "class decorator". It returns the class which inherits all the members from the original one but it modifies the `call()` method to accept data in time levels. I hope that this implementation is clear and as simple as possible. Yes, the name is the question. The `TSCoef` name, i.e. the class naming convention, is used now because it returns the class. But, maybe, `create_TSCoef` would be better, it reflects the situation more accurately.
> > 
> > 
> > You mean that you can pass the other `Coef*` in, so that those are usable for time-dependent problems? Now you can see, why an example would be nice. :) OK I get the idea, it adapts the classes for time-stepping: I would call it `adapt_for_ts()`, or something like that.
> 
> Yes, you can pass "any" other `Coef*` into it. What about `create_ts_coef`? `adapt_for_ts()`is not bad, but ...

`create_ts_coef()` is OK too. Or just `enable_ts()`?

vlukes:
Done.

rc:
Note that the appveyor tests pass too - the error is caused by some matplotlib-related output to stderr.

PR #510, opened 2019-04-24
fix material_opt.py: Python 2/3 compatibility issue - zip()

vlukes:
This fix  #509.

PR #511, opened 2019-04-26
fix material_opt.py: Python 2/3 compatibility issue - part II.

vlukes:
This commit solves #510. But logging implemented in `sfepy.base.log` seems not working in Python 3 due to some changes in multiprocessing.

### DIFF
--- a/examples/homogenization/homogenization_opt.py
+++ b/examples/homogenization/homogenization_opt.py
@@ -22,7 +22,7 @@ def get_mat(coors, mode, pb):
 
         nqp = coors.shape[0]
         nel = pb.domain.mesh.n_el
-        nqpe = nqp / nel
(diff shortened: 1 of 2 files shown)


rc:
Thanks!

BTW. Log might need some reimplementation, maybe that some tool allowing non-blocking plots appeared. Do you have any idea about the multiprocessing problem (easy/hard)?

vlukes:
On 26. 04. 19 14:18, Robert Cimrman wrote:
> Thanks!
> 
> BTW. Log might need some reimplementation, maybe that some tool allowing 
> non-blocking plots appeared. Do you have any idea about the 
> multiprocessing problem (easy/hard)?

I do not know where the snag is. I think that reimplementation may be 
faster than trying to find the multiprocessing issue.

VL

PR #512, opened 2019-04-27
Fallback solvers fixes

rc:
This fixes tests using the 'auto_direct' solver, when mpi4py is missing. @vlukes can you have a look?


vlukes:
Good! No objections.

PR #515, opened 2019-05-16
fix mumps solver for "small and dense" matrices

vlukes:


rc:
Thanks! Is there a reason for using the different names: mem_relax, mem_relaxation, memory_relaxation, mem_increase for something that seems to be a single thing? I would stick with the MUMPS notation.

vlukes:
I find MUMPS notation confusing, they call the parameter `memory_relaxation`. But in the documentation they describe it as follows: "corresponds to the percentage increase in the estimated working space". At first, I called it `mem_increase` and then changed to 'mem_relaxation'. So  I forgot to change all `mem_increase` in the code. The name of the parameter should be `memory_relaxation`  and I will use shorter `mem_relax` in the internal code. OK?

rc:
OK, for a single meaningful name in sfepy (such as mem_relax).

PR #517, opened 2019-05-22
fix VTKMeshIO.write(): incorrect tensor data shape

vlukes:


rc:
Thanks! When this happens?

vlukes:
It happens when saving non-symmetric cell data tensor with shape (nel, 1, dim, dim).

PR #519, opened 2019-05-27
Docs update citing support

rc:
@vlukes can you check the changes, please? Thanks!

### DIFF
--- a/doc/archived_support.rst
+++ b/doc/archived_support.rst
@@ -1,0 +1,29 @@
+.. only:: html
+
+   Past Support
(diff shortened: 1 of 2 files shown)


vlukes:
I suggest to use the DOI based full text link:  https://doi.org/10.1007/s10444-019-09666-0.

vlukes:
The project numbers are OK.

rc:
IMO the doi-based link shows the full-text only to people with (institutional) prescription, while the one used is for everyone.

PR #522, opened 2019-05-30
update SplineBox: cleanup, `cp_values`->`cp_coors`, ...

vlukes:


rc:
Thanks! It seems I have approved prematurely - tests/test_splinebox.py fails now - can you check it?

vlukes:
I see, something wrong in derivatives, I will check it.

vlukes:
Heh, fast writing and slow thinking. Could you discard this request? The original code is reasonable also for 1D fields, the new one is not.

PR #526, opened 2019-06-24
reimplement find_map(): use the scipy.spatial.cKDtree implementation

vlukes:


sfepy/discrete/fem/mesh.py
```diff
@@ -7,50 +7,46 @@
 from .meshio import MeshIO, supported_cell_types
 import six
 from six.moves import range
```
rc: This is now unused.

sfepy/discrete/fem/mesh.py
```diff
@@ -7,50 +7,46 @@
...
-
-    xs = xx.T[iis]
-    xd = nm.sqrt(nm.sum(nm.diff(xs, axis=0)**2.0, axis=1))
-
-    ii = nm.where(xd < eps)[0]
-    off1, off2 = ir[iis][ii], ir[iis][ii+1]
-    i1, i2 = iis[ii] - off1, iis[ii+1] - off2
-    dns = nm.where(off1 == off2)[0]
-    if dns.size:
+    x = nm.vstack([x1, x2])
```
rc: x is not used...

sfepy/discrete/fem/mesh.py
```diff
@@ -7,50 +7,46 @@
...
+    kdtree = cKDTree(nm.vstack([x1, x2]))
```
rc: x should go here.

sfepy/discrete/fem/mesh.py
```diff
@@ -7,50 +7,46 @@
...
-                output('x2: %d %d -> %s %s' % (i1[dn], i2[dn],
-                                               x2[:,i1[dn]], x2[:,i2[dn]]))
+        for dn in dns1:
+            idxs = cmap[dn, :]
+            output('x1: %d %d -> %s %s' % (idxs[0], idxs[1],
+                                           x1[idxs[0], :], x1[idxs[1], :]))
+        for dn in dns2:
+            idxs = cmap[dn, :]
+            output('x2: %d %d -> %s %s' % (idxs[0], idxs[1],
+                                           x1[idxs[0], :], x1[idxs[1], :]))
```
rc: Should not this be with x2?

rc:
Thanks, good idea! kdtree.query_pairs() seems like a useful function indeed.

I am not sure whether a global `eps` is good, though. I am aware that in periodic.py it is used as well, but nowadays I find passing this as a function parameter much cleaner.

sfepy/discrete/fem/mesh.py
vlukes: Yes, it should be.

sfepy/discrete/fem/mesh.py
vlukes: Removed

sfepy/discrete/fem/mesh.py
vlukes: Removed

vlukes:
Mesh `eps` should be controlled somehow globally. This is a possible way and I think not so bad. Now, I have no idea how to make it better. Passing it as a function argument is clean, but it is not easy to link the `eps` options to all involved functions.

rc:
Thanks for the corrections. 

It seems that the functions involved already had this argument, or you think there will be other functions?

We may come up with something better later, e.g. a context manager, or make eps an attribute of each mesh object?

vlukes:
The mesh functions have `eps` argument, but passing the mesh `eps` option defined in the problem description file to these function is not so straight forward. But, of course, somehow possible.

Yes, maybe make `eps` a mesh object attribute is the solution.

rc:
OK, let us merge this as it is now.

rc:
OK, let us merge this as it is now. BTW. the new problem configuration option(s) should be documented in http://sfepy.org/doc-devel/users_guide.html#miscellaneous.

PR #527, opened 2019-06-24
new remote mirror regions: allow mirror regions with no common vertices

vlukes:
You can define _remote_ mirror region as:  `regions = {'XX': ('...', 'facet', 'AA',  {'mirror_region': 'YY'})}`, such that `XX` a `YY` have no common vertices, but coors_XX = coors_YY  + C.

### DIFF
--- a/sfepy/discrete/common/region.py
+++ b/sfepy/discrete/common/region.py
@@ -676,19 +676,58 @@ class Region(Struct):
         """
         Find the corresponding mirror region, set up element mapping.
         """
+        from sfepy.discrete.fem.mesh import find_map
+
         regions = self.domain.regions
+        eopts = self.extra_options
 
-        for reg in regions:
-            mirror_parent = regions.find(reg.parent)
-            if mirror_parent is None: continue
-            if ((reg is not self)
-                and nm.all(self.vertices == reg.vertices)):
-                mirror_region = reg
-                break
+        if self.mirror_region is not None:
+            return
+
+        if (eopts is not None) and ('mirror_region' in eopts):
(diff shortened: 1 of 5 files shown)


sfepy/discrete/common/region.py
```diff
@@ -676,19 +676,62 @@ def setup_mirror_region(self):
...
+                / coors1.shape[0])
+            vmap1, vmap2 = find_map(coors1, coors2 - shift, join=False)
+            if vmap1.shape[0] != coors1.shape[0]:
+                print(coors1[vmap1])
+                print(coors2[vmap2])
+                raise ValueError('cannot match vertices!')
+
+            fconn = self.domain.cmesh.get_conn_as_graph(self.dim - 1, 0)
+            # region facet centers
+            fcc = nm.empty((self.facets.shape[0], self.dim), dtype=nm.float64)
```
rc: Maybe just use CMesh.get_centroids() which computes all facet centroids? I do not think the overhead would be that much. Anyway, get_centroids(region) is generally useful and should be available as a function.

rc:
Thanks for this new feature! In what kind of problems you need that? It looks like some kind of periodic conditions? (Could it not be simulated by those?)

sfepy/discrete/common/region.py
vlukes: OK, I did not remember this function.

vlukes:
It is intended for traces on various "non-collocated" interfaces. See the acoustic example, where the interacting parts are glued.

rc:
OK, but the feature is not used/tested in sfepy examples, right? Would it be easy to modify one to use it?

PR #528, opened 2019-06-25
Mirror region misc

vlukes:
Explanation to `get_conn_key()`:

consider the following regions:
```
'Surf1': ('...', 'facet', 'Y', {'mirror_region': 'Surf2'}),
'Surf2': ('...', 'facet', 'Y', {'mirror_region': 'Surf1'}),
```

and two variables `p` a `v`, where the field of the virtual variable `v` is a surface field defined at `Surf1` and 'p' may be the variable of a volume field defined in region `Y`, which includes `Surf1` and `Surf2`. Than the proposed fix is required to distinguish between: 
- `dw_surface_dot.i1.Surf1(v, p)` - integration over `Surf1`, where `v` and `p` are defined 
- `dw_surface_dot.i1.Surf1(v, tr(p))` - integration over `Surf1` with the trace of `p` on `Surf2`

### DIFF
--- a/doc/users_guide.rst
+++ b/doc/users_guide.rst
@@ -1095,6 +1095,14 @@ Additional options (including solver selection)::
 
         # string, a function to modify problem definition parameters
         'parametric_hook' : '<parametric_hook_function>',
+
+        # float, default: 1e-9. If the distance between two mesh vertices
+        # is less than this value, they are considered identical.
+        # This affects:
+        # - periodic regions matching
+        # - mirror regions matching
+        # - fixing of mesh doubled vertices
+        'mesh_eps': 1e-7,
     }
 
 * ``post_process_hook`` enables computing derived quantities, like
--- a/doc/users_guide.rst
+++ b/doc/users_guide.rst
@@ -541,7 +541,7 @@ Region Definition Syntax
 Regions are defined by the following Python dictionary::
 
         regions = {
-            <name> : (<selection>, [<kind>], [<parent>]),
+            <name> : (<selection>, [<kind>], [<parent>], [{<misc. options>}]),
         }
 
 or::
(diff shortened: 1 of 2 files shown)


rc:
Thanks for the docs update!

> Explanation to `get_conn_key()`:
> 
> consider the following regions:
> 
> ```
> 'Surf1': ('...', 'facet', 'Y', {'mirror_region': 'Surf2'}),
> 'Surf2': ('...', 'facet', 'Y', {'mirror_region': 'Surf1'}),
> ```
> 
> and two variables `p` a `v`, where the field of the virtual variable `v` is a surface field defined at `Surf1` and 'p' may be the variable of a volume field defined in region `Y`, which includes `Surf1` and `Surf2`. Than the proposed fix is required to distinguish between:
> 
>     * `dw_surface_dot.i1.Surf1(v, p)` - integration over `Surf1`, where `v` and `p` are defined
> 
>     * `dw_surface_dot.i1.Surf1(v, tr(p))` - integration over `Surf1` with the trace of `p` on `Surf2`

This seems a bit counter-intuitive to me, but I understand it. `tr(x)` for us means "take `x` values from the mirror region of the term region" - maybe we need a new notation. Maybe specify the mirror region name explicitly in the equation definition by using `tr(Surf2, p`) or something similar?

BTW. I have merged #527, so you may want to rebase on master.

vlukes:
This `tr(Surf2, p)` would be perfect, but I'm not to familiar with parsing terms.

vlukes:
Rebased

rc:
> This `tr(Surf2, p)` would be perfect, but I'm not to familiar with parsing terms.

OK, let us merge this as it is now, and I will try to enrich the term syntax later. (A fully symbolic was of building the equations would be even better, but out of scope now.)

rc:
BTW. it would be good to have an example for playing with the `tr` syntax.

PR #530, opened 2019-06-26
update region.setup_mirror_region() for `tr(reg, var)` syntax

vlukes:
It is here.


rc:
Something broke in the CI configuration (it reports missing atlas), so let's merge this and fix the CI issues later.

PR #532, opened 2019-06-28
fix traces for evaluation mode

vlukes:
I hope nothing breaks. All tests pass.


rc:
Thanks! The trace code needs more tests to catch such things. Maybe adding an example using those features would be sufficient for now?

PR #533, opened 2019-07-01
CI update

BubuLK:
Updated existing CI config files to follow new conda-4.7.x behavior.

Notes:
- as part of PR python 3.6->3.7 config changes was made, 
- there are some known problems with latest conda command on Windows, so update was temorary disabled (but still version 4.7 is used),
- there are still no CI config for MacOS platform ;(
- some additional clean-up of config files needed in (near) future...

LK


rc:
Thanks!

PR #545, opened 2019-08-12
fix ls.schurs_mumps: 'bloc' -> 'block', use active dof info

### LINKED ISSUE #544: ls.schur_mumps segfaults with Python 3.7, MUMPS 5.1.2
It happens on line 430 of `sfepy/solvers/ls_mumps.py`, when running
```
python3 simple.py examples/linear_elasticity/linear_elastic_up.py
```

vlukes:
This should fix #544.


rc:
Thanks!

PR #556, opened 2019-09-11
update convert_mesh.py: new '23d' option, write shell cells only

vlukes:


rc:
What about adding instead a `--cell-dim` option, that could take a list of ints from {1, 2, 3}? This could replace the `-23` as well as `-3`. The `-2` is different and would stay (it is used for fixing planar files with zero/constant z coordinate).

So:
`-3` -> `--cell-dim=3`
`-23` -> `--cell-dim=2`

I would allow also  `--cell-dim=2,3` for meshes to be used outside sfepy.

vlukes:
OK, `--cell-dim` (short version is  `-d`) option added. Also syntax `--cell-dim=2,3`is allowed.

sfepy/discrete/fem/mesh.py
```diff
@@ -386,7 +386,11 @@ def _get_io_data(self, cell_dim_only=None):
         cmesh = self.cmesh
         conns, mat_ids = [], []
         if cell_dim_only is not None:
-            descs = [ii for ii in self.descs if int(ii[0]) == cell_dim_only]
+            try:
+                iter(cell_dim_only)
```
rc: Maybe use `is_sequence()` from sfepy.base.base

script/convert_mesh.py
```diff
@@ -39,7 +39,7 @@
     'tri-tetra' : 'convert elements: quad->tri, hexa->tetra',
     '2d' : 'force a 2D mesh by removing the z coordinates - assumes a 3D mesh'
     ' in the xy plane',
-    '3d' : 'force a 3D mesh by removing 1D and 2D cells',
+    'cell-dim' : 'write only cells of a given dimension(s)',
```
rc: Maybe more explicitly:
"write only cells of a given dimension - use a comma-separated list for several values."

rc:
Thanks! I have two small suggestions, otherwise LGTM.

script/convert_mesh.py
vlukes: updated

rc:
`is_sequence()` does not work?

sfepy/discrete/fem/mesh.py
vlukes: It works but I'm not so fast.

rc:
Thanks, merging!

PR #559, opened 2019-09-24
Misc updates

zitkat:
Added simple decorator for convenience.

### DIFF
--- a/sfepy/discrete/functions.py
+++ b/sfepy/discrete/functions.py
@@ -48,25 +48,26 @@ def make_sfepy_function(fun_or_name=None):
     Convenience decorator to quickly create
     sfepy.discrete.functions.Function objects
 
-    Has two modes of use either without parameter:
+    Has two modes of use either without parameter::
 
         @make_sfepy_function
         def my_function(...):
             ...
 
-        or with name
+    or with name::
 
         @make_sfepy_function("name_for_my_function")
         def my_function(...):
             ...
+
     Parameters
     ----------
-    fun_or_name: string, optional
+    fun_or_name : string, optional
         name to be saved within Function object, if None name of decorated function is used
 
     Returns
     -------
-    sfepy_function : sfepy.discrete.functions.Function
+    sfepy.discrete.functions.Function
         with attribute name set to provided name or original function name
     """
     if callable(fun_or_name):
@@ -78,11 +79,11 @@ def make_sfepy_function(fun_or_name=None):
 
         Parameters
         ----------
-        fun: callable to be converted to sfepy.siscrete.functions.Function
+        fun : callable to be converted to sfepy.discrete.functions.Function
 
         Returns
         -------
-        sfepy.siscrete.functions.Function object
+        sfepy.siscrete.functions.Function
         """
         if fun_or_name is not None:
             return Function(fun_or_name, fun)
--- a/sfepy/discrete/functions.py
+++ b/sfepy/discrete/functions.py
@@ -48,10 +48,26 @@ def make_sfepy_function(fun_or_name=None):
     Convenience decorator to quickly create
     sfepy.discrete.functions.Function objects
 
-    Can be used directly without parameter.
+    Has two modes of use either without parameter:
 
-    :param fun_or_name: name saved within Function object, if None name of decorated function is used
-    :return:
+        @make_sfepy_function
+        def my_function(...):
+            ...
+
+        or with name
+
+        @make_sfepy_function("name_for_my_function")
+        def my_function(...):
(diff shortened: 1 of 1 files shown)


sfepy/discrete/functions.py
```diff
@@ -42,6 +43,52 @@ def set_function(self, function, is_constant=False):
     def set_extra_args(self, **extra_args):
         self.extra_args = extra_args
 
+def make_sfepy_function(fun_or_name=None):
+    """
+    Convenience decorator to quickly create
+    sfepy.discrete.functions.Function objects
+
+    Has two modes of use either without parameter:
```
rc: Use :: instead of : to make the code block rendered properly:
```
It has two modes of use, either with or without a parameter::
```

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,52 @@ def set_function(self, function, is_constant=False):
...
+    if callable(fun_or_name):
+        return Function(fun_or_name.__name__, fun_or_name)
+
+    def functionizer(fun):
+        """
+        Internal decorator
+
+        Parameters
+        ----------
+        fun: callable to be converted to sfepy.siscrete.functions.Function
```
rc: siscrete -> discrete

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,52 @@ def set_function(self, function, is_constant=False):
...
+
+        Returns
+        -------
+        sfepy.siscrete.functions.Function object
```
rc: see above

rc:
Thanks! I will have a look at it soon, for now only the comments above.

sfepy/discrete/functions.py
zitkat: Fixed, also added :: to the later comment line or with name:: to get two block.

sfepy/discrete/functions.py
zitkat: Fixed it.

sfepy/discrete/functions.py
zitkat: Fixed it.

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,54 @@ def set_function(self, function, is_constant=False):
...
+
+    Parameters
+    ----------
+    fun_or_name : string, optional
+        Name to be saved within `Function` instance,
+        if None name of decorated function is used.
+
+    Returns
+    -------
+    `sfepy.discrete.functions.Function`
```
rc:
```
new_fun : Function instance
    The Function object.
```

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,54 @@ def set_function(self, function, is_constant=False):
...
+        """
+        Internal decorator
+
+        Parameters
+        ----------
+        fun : callable to be converted to sfepy.discrete.functions.Function
+
+        Returns
+        -------
+        sfepy.discrete.functions.Function
```
rc: Maybe not necessary?

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,54 @@ def set_function(self, function, is_constant=False):
...
+        Name to be saved within `Function` instance,
+        if None name of decorated function is used.
+
+    Returns
+    -------
+    `sfepy.discrete.functions.Function`
+        With attribute name set to provided name or original function name.
+    """
+    if callable(fun_or_name):
+        return Function(fun_or_name.__name__, fun_or_name)
```
rc: Why is this branch actually needed? Is it not covered by `return Function(fun.__name__, fun)` in `functionizer()`?

zitkat: This allows usage where you omit not only parameters but also parenthesis and use decorator like in the first mode of use in documentation commentary:
```python
@make_sfepy_function
def my_function(...):
            ...
```

rc: I see, thanks!

sfepy/discrete/functions.py
```diff
@@ -42,6 +43,57 @@ def set_function(self, function, is_constant=False):
...
+        Parameters
+        ----------
+        fun : callable
+         Will be converted to sfepy.discrete.functions.Function.
+
+        Returns
+        -------
+        new_fun : Function instance
+            The `sfepy.discrete.functions.Function` object.
+
```
rc: Remove space? BTW. you can amend the previous commit and force-push again ;)

rc:
Thanks, I have two last nit-picking comments (see above), otherwise LGTM!

PR #560, opened 2019-09-24
Gmsh write

zitkat:
Added read_data and write to Msh2MeshIO for partial support of GMSH 2.0 format, according to http://gmsh.info/doc/texinfo/gmsh.html#File-formats

### DIFF
--- a/sfepy/discrete/fem/meshio.py
+++ b/sfepy/discrete/fem/meshio.py
@@ -2897,11 +2897,11 @@ class Msh2MeshIO(MeshIO):
 
     def write(self, filename, mesh, out=None, ts=None, **kwargs):
         """
-
+        Writes data into msh v2.0 file, handles cell_nodes data from DGField
         :param filename: path to file
         :param mesh: computational mesh
         :param out: data on the computational mesh
-        :param ts: time step?
+        :param ts: time stepper?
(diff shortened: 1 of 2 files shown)


rc:
Thanks! Could you rebase on master so that the merge commits disappear? I can help with that, if needed.

sfepy/discrete/fem/meshio.py
```diff
@@ -2822,7 +2902,7 @@ def read(self, mesh, omit_facets=True, **kwargs):
                 pass
 
             else:
-                output('skipping unknown entity: %s' % line)
+                # output('skipping unknown entity: %s' % line)
```
rc: This line can be simply removed. The other readers do not print such information. (Alternative: add `verbose` argument to `read()`.)

zitkat:
> Thanks! Could you rebase on master so that the merge commits disappear? I can help with that, if needed.

I don't see any merge commits, I did rebase just after 34b0110 so maybe it resolved them.

sfepy/discrete/fem/meshio.py
zitkat: I removed it.

rc:
Yes, I hoped rebase would remove the merge commits, thanks.

PR #571, opened 2020-01-14
fix `dw_surface_dot` term: nvec orient. - outwards to the parent regi…

vlukes:
…on of the virtual variable


rc:
Thanks, LGTM! Could you fetch the current master and rebase upon it?

vlukes:
Rebased.

PR #575, opened 2020-01-20
fix argument type in SplineRegion2D.create_spb()

rc:
- fixes Travis CI failure for Python 3.7, numpy-1.18.1

@vlukes, do you agree with the fix? See #574 for the failure.


vlukes:
I agree, `rho` must be the integer type.

rc:
Thanks, I will merge it.

PR #579, opened 2020-02-05
drop python 2.7 support

rc:
@vlukes Could you check this PR?


vlukes:
Yes, Python 2.7 can be dropped. I do not use PysparseEigenvalueSolver, so I will not miss it.

rc:
OK, let us merge it then.

PR #581, opened 2020-02-17
force dtype in {CoefNN, CoefN, CoefOne}.set_variables_default()

rc:
Fixes numpy casting TypeError when adding inplace a complex128 array to a float64 one.

@vlukes, do you agree? Another possibility is to replace in-place addition by the regular one, but that creates temporary arrays.


vlukes:
:thumbsup: :wink:

rc:
Thanks, will merge it.

PR #583, opened 2020-03-06
fix saving correctors in OnesDim

rc:
Hi @vlukes, could you check this? Thanks!


vlukes:
It is okey. It should not break anything in my code.

Issue #303, opened 2016-02-09
new logo?


rc:
Two proposals for logos with snakes created by Serge-Étienne Parent:

![rect6236](https://cloud.githubusercontent.com/assets/44282/12918490/bceaac74-cf3f-11e5-87ca-248c224a23fc.png)

![rect6240](https://cloud.githubusercontent.com/assets/44282/12918480/b5065530-cf3f-11e5-8ec7-ad3bc79386df.png)

Comments and suggestions welcome!
r.

BubuLK:
+1 for second version. Maybe in variant with FEM-like colors (oposit to not-so-easy-understand NTC-like colors).

essicolo:
What do you mean by FEM-like colors?

rc:
On 02/10/2016 05:14 PM, Serge-Étienne Parent wrote:
> What do you mean by FEM-like colors?

Probably coming from a problem solution on a snake-like mesh. The current logo 
is a view of a 3D Poisson problem solution on a cube.

BubuLK:
Many thanks to Robert for explanation ;) and sorry for confusion :(

Please, let me a little bit to clarify my 'shortcut' comment:

  - I really like both logo proposals!
  - BTW: I'm ITC guy and don't really understand all the SfePy/FEM
    math/physics 'magic' :((
  - but...I really appreciate Robert's (and his co-workers)
    effort and really like his Python coding approach :)
  - I'm BIG fan of good typeseting/logos/presentation (including good old Knuth TeX style) 

As far as I understand, current logo proposal is base on
color scheme derived from 'visual style' of
'New Technologies Research Center of University of West Bohemia'
(aka NTC in Czech language, sorry :()
BTW: I'm not big fan of violet colors...

So, my proposal was to create alternative colors scheme (as explained
by Robert) for the Python (or part of them).
Of course, I can't imagine final results and understand possible
problems with printing such colorful logo.

Sorry for my really bad English and hopefully you understand
what I'm trying to explain :)

rc:
On 02/10/2016 10:58 PM, Lubos Kejzlar wrote:
> Many thanks to Robert for explanation ;) and sorry for confusion :(

No need to apologize :)

> Please, let me a little bit to clarify my 'shortcut' comment:
>
>    - I really like both logo proposals!
>    - BTW: I'm ITC guy and don't really understand all the SfePy/FEM
>      math/physics 'magic' :((
>    - but...I really appreciate Robert's (and his co-workers)
[...]

PR #588, opened 2020-03-26
New SfePy web

vlukes:
The big update of SfePy we pages:

- sphinx "Read the Docs" theme is used, provides modern look and optimization for mobile devices
- drop own download page, use github instead
- rearrange main menu
- new sfepy logo, the one proposed by @essicolo, see #303 (not yet in the final version)
- a little cleaning


rc:
Thank you, great!

I have trouble building the docs:
```
ImportError: No module named 'sphinx_rtd_theme'
```
Maybe add the file into git?

vlukes:
`sphinx_rtd_theme` is the python module, install it with `pip`:
```
pip install sphinx_rtd_theme
```

rc:
Thanks, that helped!

I have a few remarks:
- The term table is not fully visible, and it is worse with scripts enabled. The problem might be related to the fact, that `Ctrl--` does not only decrease the font size, but also decreases the width of the menu and text columns. The middle table column is very wide and the equation column very narrow. I missed the table slider at the bottom when I looked the first time. The original (fixed, no scrollbar) table layout seems preferable?
- The directory structure table in the developer guide is also too wide (the second column), but that is not so important.
- The [gallery](http://sfepy.org/gallery/) link is missing - I would add it to the panel menu.
- The docstrings are in numpy doc format, and the rendering of parameters is not good. For example, `pars : array_like` is rendered as  `parsarray_like`.

Other sections seem OK, well done!

The sphinx_rtd_theme needs to be either included in `doc/`, or mentioned in [requirements](http://sfepy.org/doc-devel/installation.html#requirements) as a documentation build dependency.

vlukes:
- The issue with the tables is fixed.
- The gallery link added to the menu (`Documentation`).
- Read the Docs theme added to requirements.
- The menu depth of `Documentation` set to 2.
- Can not reproduce the last reported issue.

rc:
Thanks, the tables look great now!

As for the last issue, I have Sphinx v2.4.4, sphinx_rtd_theme 1.12.0, numpydoc 0.9.2, the result is in the attached figure. What is your numpydoc version?

Screenshot_20200328_235351

More notes:

- Another issue appeared, maybe related to the documentation menu depth - the table of contents in the PDF manual contains only the top chapters, which is insufficient. Why was the depth limit needed? Or you want to drop the PDF manual (the link to it was removed)? IMO it is a useful thing to have.

- The table of contents in the developer guide with backlinks to top was quite handy - can it be brought back?

- I would put also back git commit ID to the version. We could use only a short id, as in `git rev-parse --short HEAD` - i.e. only the first 8 characters.

- Maybe put a link to the gallery to the front page? (`check Examples and Tutorial` -> `check Gallery, Examples and Tutorial`?

vlukes:
- My Sphinx version is 1.6.7 and numpydoc is 0.7.0.
- I have reduced the Documentation menu depth because it was too long.
- Yes, I was thinking that the PDF documentation could be dropped. Do we really need it?
- I do not know how to do the backlinks in the developer guide in the current arrangement. The menu or the development and dev. guide pages would have to be rearranged to make it work. 
- OK, the short id is right.
- Yes, I can put the link to the gallery as you proposed.

rc:
Downgrading to numpydoc 0.7.0 solved the docstring issue, but your sphinx fix helped too, so numpydoc 0.9.2 works, thanks!

script/gen_gallery.py
```diff
@@ -501,216 +487,200 @@ def generate_rst_files(rst_dir, examples_dir, images_dir):
...
-        ifd = open(os.path.join(full_dirname, 'index.rst'), 'w')
-        ifd.write(_index % (dirname, dirname, '=' * len(dirname)))
+        ifd = open(os.path.join(rst_dir, '%s-index.rst' % dirname), 'w')
+        ifd.write(_index % (dirname + '-examples',
+                            dirname, '=' * len(dirname)))
 
         for ex_filename, rst_filename in filenames:
-            full_rst_filename = os.path.join(full_dirname, rst_filename)
-            output('"%s"' % full_rst_filename.replace(rst_dir, '')[1:])
-            rst_filename_ns = rst_filename.replace('.rst', '')
```
rc, line 528: This is not safe as it replaces all occurrences of '.rst'. Use `os.path.splitext()`, please. (Also at other places in this file.) Or, better, use `edit_filename()` from `sfepy.base.ioutils`.

PR #589, opened 2020-04-01
Fix import from meshio

heczis:
VTKMeshIO no longer exists (among other similar classes in meshio), MeshioLibIO seems to do the job here.


rc:
Thanks, LGTM!

PR #593, opened 2020-04-10
Gmshio thesecond

zitkat:
Basically the same class as in my previous pull request, this time read and writing of mesh and vertex and cell data is done by meshio library. I used class MeshioLibIO as base class and included some minor refactoring of it.  Please note my comments regarding `"%s:ref".self.file_format` data, they cause some trouble in gmsh as they overlay other data in visualization and when loading multiple files e.g. multiple time steps the multiple of them is loaded. For now Gmshio excludes them.

### DIFF
--- a/sfepy/discrete/fem/meshio.py
+++ b/sfepy/discrete/fem/meshio.py
@@ -29,12 +29,16 @@ _supported_formats = {
     # modes: r = read, w = write, c = test cell groups, v = test vertex groups
     'abaqus': ('meshio', None, 'cv'),
     'exodus': ('meshio', None, 'v'),
-    # 'gmsh': ('meshio', None, 'cv', ['gmsh4-binary', 'gmsh4-ascii',
-    #                                 'gmsh2-binary', 'gmsh2-ascii']),
-    'gmsh': ('gmshio', ".msh", 'rw'),
+    # If file_format is not provided explicitly: in Python 3.5, the dict order
+    # is not guaranteed, so a gmsh file can be read, either by 'meshio' or
+    # 'gmshio' in a random way, in Python 3.6+, meshio is always preferred.
+    # Variants do not work with the exception of 'gmsh-dg'.
+    'gmsh': ('meshio', None, 'cv', ['gmsh4-binary', 'gmsh4-ascii',
+                                    'gmsh2-binary', 'gmsh2-ascii']),
+    'gmsh-dg': ('gmshio', '.msh', 'rw', ['gmsh-dg']),
     'medit': ('meshio', None, 'cv'),
     'nastran': ('meshio', None, 'cv'),
-    'vtk': ('meshio', None, 'cv', ['vtk-binnary', 'vtk-ascii']),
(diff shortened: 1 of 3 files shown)


rc:
@zitkat, I have rebased this PR branch onto master and added format_variant application option, see [my fork](https://github.com/rc/sfepy/commits/gmshio-thesecond). I needed to make some changes to both `MeshioLibIO` and `GmshIO` to make it work together. Some issues remain, @vlukes, would you mind having a look?

FWIW, I tested the changes by adding the following to the `options` in `examples/diffusion/poisson.py`:
```
    'output_format' : 'msh',
    'format_variant' : 'gmsh-dg',
    # 'output_format' : 'vtk',
    # 'format_variant' : 'vtk-ascii',
```

zitkat:
I have included your changes into branch on my fork. It works fine on DG examples I have tested. Note that the format variant can be used with `Problem.save_state` as well, in that case it has to be passed as `file_format`:
```python
pb.save_state("sol-h16o01.0.msh", state=pb.sol, file_format="gmsh-dg")
```
This works fine with me, but could be confusing and took some time work out.

rc:
Yes, it is not ideal. The "right" thing to do would be to use the existing `'output_format'` option, but that option is suffix-based, and is "everywhere" (code, docstrings, docs...) - I was not willing to do the tedious change in this PR ;). I thought about using `'file_format'` for the option and found it too similar to `'output_format'`, but it might be better after all.

rc:
@zitkat, I have just renamed the option to `file_format`, you can fetch it from [my fork](https://github.com/rc/sfepy/commits/gmshio-thesecond).

sfepy/discrete/fem/meshio.py
```diff
@@ -438,18 +450,23 @@ def write(self, filename, mesh, out=None, ts=None, **kwargs):
...
     def read_data(self, step, filename=None, cache=None):
-        m = meshiolib.read(self.filename, file_format=self.file_format)
+        """
+        Renames cell resp. vertex data with name "*:ref"
+        to mat_id resp. node_groups
+
+        :param step: has no effect
+        :param filename: filename to use insted of self.filname
+        :param cache: has no effect
+        :return:
```
rc, line 465: Could you convert the docstring to the numpydoc format? Thanks!

zitkat: Missed that, done.

sfepy/discrete/fem/meshio.py
```diff
@@ -438,18 +450,31 @@ def write(self, filename, mesh, out=None, ts=None, **kwargs):
...
     def read_data(self, step, filename=None, cache=None):
-        m = meshiolib.read(self.filename, file_format=self.file_format)
+        """
+        Renames cell resp. vertex data with name "*:ref"
+        to mat_id resp. node_groups
+
+        Parameters
+        ----------
+        step: has no effect
+        filename: filename to use instead of self.filname
```
rc, line 465: Thanks, just one typo (self.filname), and maybe use the following:
```
filename : string, optional
    The file name to use instead of self.filename.
```

PR #594, opened 2020-04-10
Dg sfepy integration independent

zitkat:
This branch contains selected changes I made to the sfepy code during DG developement. I selected only changes that do not require any DG specific files, as you can see there is not many of them. I will also add changes which ensure parsing of DG boundary conditions lated - these can be integrated without DG files - however I have to resolve failing tests before that.

### DIFF
--- a/sfepy/discrete/common/extmods/cmesh.pyx
+++ b/sfepy/discrete/common/extmods/cmesh.pyx
@@ -688,7 +688,12 @@ cdef class CMesh:
         ccf = self.get_conn(td, td - 1)
 
         out = np.empty((ccf.n_incident, self.dim), dtype=np.float64)
-        mesh_get_facet_normals(self.mesh, &out[0, 0], which)
+        if self.dim == 1:
+            # TODO quick fix for normals, relies on nice ordering of 1D mesh
+            # i.e. cell facets are indexed from left to right
+            out = np.tile(np.array([-1, 1], dtype=np.float64), ccf.n_incident//2)[:, None]
+        else:
+            mesh_get_facet_normals(self.mesh, &out[0, 0], which)
 
         return out
 
--- a/sfepy/discrete/common/extmods/cmesh.pyx
+++ b/sfepy/discrete/common/extmods/cmesh.pyx
@@ -691,7 +691,7 @@ cdef class CMesh:
         if self.dim == 1:
             # fix for 1D normals, relies on nice ordering of 1D mesh
             # i.e. cell facets are uniformly indexed from left to right
-            # globaly and localy
+            # globally and locally
             out = np.tile(np.array([-1, 1], dtype=np.float64),
                           ccf.n_incident//2)[:, None]
         else:
--- a/sfepy/discrete/common/extmods/cmesh.pyx
+++ b/sfepy/discrete/common/extmods/cmesh.pyx
@@ -689,9 +689,11 @@ cdef class CMesh:
 
         out = np.empty((ccf.n_incident, self.dim), dtype=np.float64)
         if self.dim == 1:
-            # TODO quick fix for normals, relies on nice ordering of 1D mesh
-            # i.e. cell facets are indexed from left to right
-            out = np.tile(np.array([-1, 1], dtype=np.float64), ccf.n_incident//2)[:, None]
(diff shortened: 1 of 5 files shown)


sfepy/discrete/common/extmods/cmesh.pyx
```diff
@@ -688,7 +688,12 @@ cdef class CMesh:
         ccf = self.get_conn(td, td - 1)
 
         out = np.empty((ccf.n_incident, self.dim), dtype=np.float64)
-        mesh_get_facet_normals(self.mesh, &out[0, 0], which)
+        if self.dim == 1:
+            # TODO quick fix for normals, relies on nice ordering of 1D mesh
```
rc, line 692: Is it still TODO? :) See also fb3c3d476e10562a821a7c0594fd87080c35716a - something similar was done for SurfaceMapping.

sfepy/discrete/fem/refine.py
```diff
@@ -7,6 +7,33 @@
...
+    new_coors[0::2] = mesh_in.coors
+    new_coors[1::2] = c_centres
+
+    n_nod = len(new_coors)
+
+    new_conn = nm.arange(n_nod, dtype=nm.int32).repeat(2)[1:-1].reshape((-1, 2))
+
+    if not nm.all(cmesh.cell_groups[0] == cmesh.cell_groups):
+        raise Warning("Multiple mat ids not supported when refining, using cmesh.cell_groups[0] == {}".format(cmesh.cell_groups[0]))
+    new_mat_id = nm.zeros(n_nod - 1, dtype=nm.int32) + cmesh.cell_groups[0]
```
rc, line 31: It should be possible to preserve the cell groups like in the other refine functions.

sfepy/discrete/problem.py
```diff
@@ -621,7 +621,8 @@ def update_equations(self, ts=None, ebcs=None, epbcs=None,
         graph_changed = self.equations.time_update(self.ts,
                                                    ebcs, epbcs, lcbcs,
                                                    functions, self,
-                                                   active_only=ac)
+                                                   active_only=ac,
+                                                   verbose=get_default_attr(self.conf, 'verbose', False))
```
rc, line 625: Do not change the default, please. You can use `self.conf.get('verbose', True)`. The same applies to `self.update_materials()` calls below.

rc:
Thanks!

I made only a few small comments, otherwise it looks good. (I did not check possible whitespace errors or line lenghts, assuming your IDE does that.)

sfepy/discrete/common/extmods/cmesh.pyx
zitkat: I have checked fb3c3d4 however it does not work for me.  I'm not using SurfaceMapping in DG only VolumeMapping so maybe I missed something, but it does not return normals for each face but for each qp, which is not what I need. Also wouldn't it be nice if cmesh returned correct normals even for 1D?

sfepy/discrete/fem/refine.py
zitkat: Ok it was actually quite simple.

sfepy/discrete/common/extmods/cmesh.pyx
rc: Yes, I am aware that it does not help your case - it was just to show that a similar "hack" (special case overwriting in Python) had been used before.

In 1D, the surface is a single point, so the C function cannot guess the correct direction - the information is missing. So, yes, it would be nice, but it is not worth the effort with the current C code/data structures.

sfepy/discrete/common/extmods/cmesh.pyx
```diff
@@ -688,7 +688,14 @@ cdef class CMesh:
         ccf = self.get_conn(td, td - 1)
 
         out = np.empty((ccf.n_incident, self.dim), dtype=np.float64)
-        mesh_get_facet_normals(self.mesh, &out[0, 0], which)
+        if self.dim == 1:
+            # fix for 1D normals, relies on nice ordering of 1D mesh
+            # i.e. cell facets are uniformly indexed from left to right
+            # globaly and localy
```
rc, line 694: globally, locally

rc:
Thanks for the update, I just found two typos...

rc:
LGTM!

PR #595, opened 2020-04-12
lower memory usage of HomogenizationEngine

rc:
Hi @vlukes,

I have tried forcing garbage collection in `HomogenizationWorker.__call__()` and the results are (on one of my problems):

memory_profiles

The black curve is before and the blue one is after the patch. As you can see, the improvement is significant, while the time overhead is next to none (the time steps coincide in the figure).

I wonder if the same thing can be done in the parallel workers - do you have an idea (patches welcome), where to put the `gc.collect()` calls? It would be great if you could try it on some of your large problems that you run in parallel.

For the profiling, I am using https://github.com/pythonprofilers/memory_profiler:
```
mprof run <script.py>
mprof plot
```

### DIFF
--- a/sfepy/homogenization/engine.py
+++ b/sfepy/homogenization/engine.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, division
+import gc
 from copy import copy
 
 from sfepy.base.base import output, get_default, Struct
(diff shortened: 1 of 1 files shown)


vlukes:
Hi @rc,

your reduction of memory usage is really nice! Only two new lines and such improvements! In the near future, I will need to solve problems of homogenization of large deforming structures, so I will try to experiment with garbage collection as you propose.

rc:
OK, let's merge this as it is now, and if you find some other good places for gc, it will go to another PR.

PR #596, opened 2020-04-12
DG sfepy bc integration

zitkat:
Here are changes to process DG specific boundary conditions. It might make more sense to merge them with the rest of DG but this way the code i comming in smaller chunks.

### DIFF
--- a/sfepy/base/conf.py
+++ b/sfepy/base/conf.py
@@ -22,7 +22,8 @@ _required = ['filename_mesh|filename_domain', 'field_[0-9]+|fields',
              'region_[0-9]+|regions', 'variable_[0-9]+|variables',
              'material_[0-9]+|materials',
              'solver_[0-9]+|solvers']
-_other = ['epbc_[0-9]+|epbcs', 'lcbc_[0-9]+|lcbcs', 'nbc_[0-9]+|nbcs',
+_other = ['dgebc_[0-9]+|dgebcs', 'epbc_[0-9]+|epbcs', 'dgepbc_[0-9]+|dgepbcs',
+          'lcbc_[0-9]+|lcbcs', 'nbc_[0-9]+|nbcs',
           'ic_[0-9]+|ics', 'function_[0-9]+|functions', 'options',
           'integral_[0-9]+|integrals']
 
@@ -84,6 +85,9 @@ def transform_conditions(adict, prefix):
 
     return d2
 
+def transform_dgebcs(adict):
+    return transform_conditions(adict, "dgebcs")
+
 def transform_ebcs(adict):
     return transform_conditions(adict, 'ebc')
 
@@ -120,7 +124,7 @@ def transform_lcbcs(adict):
 
     return d2
 
-def transform_epbcs(adict):
+def transform_epbcs(adict, prefix="epbc"):
     d2 = {}
     for ii, (key, conf) in enumerate(six.iteritems(adict)):
         if isinstance(conf, tuple):
@@ -131,11 +135,14 @@ def transform_epbcs(adict):
                 c2 = tuple_to_conf(key, conf,
                                    ['region', 'times', 'dofs', 'match'])
 
-            d2['epbc_%s__%d' % (c2.name, ii)] = c2
+            d2['%s_%s__%d' % (prefix, c2.name, ii)] = c2
         else:
             c2 = transform_to_struct_1(conf)
-            d2['epbc_%s' % c2.name] = c2
+            d2['%s_%s' % (prefix, c2.name)] = c2
     return d2
+
+def transform_dgepbcs(adict):
+    return transform_epbcs(adict, "dgepbc")
 
 def transform_regions(adict):
     d2 = {}
@@ -261,6 +268,8 @@ transforms = {
     'variables' : transform_variables,
     'ebcs'      : transform_ebcs,
     'epbcs'     : transform_epbcs,
+    'dgebcs'    : transform_dgebcs,
+    'dgepbcs'   : transform_dgepbcs,
     'nbcs'      : transform_to_struct_01,
     'lcbcs'     : transform_lcbcs,
     'ics'       : transform_ics,
--- a/sfepy/base/conf.py
+++ b/sfepy/base/conf.py
@@ -86,7 +86,7 @@ def transform_conditions(adict, prefix):
     return d2
 
 def transform_dgebcs(adict):
-    return transform_conditions(adict, "dgebcs")
+    return transform_conditions(adict, "dgebc")
 
 def transform_ebcs(adict):
     return transform_conditions(adict, 'ebc')
--- a/sfepy/discrete/common/dof_info.py
+++ b/sfepy/discrete/common/dof_info.py
@@ -272,6 +272,15 @@ class EquationMap(Struct):
         self.dpn = len(self.dof_names)
         self.eq = nm.arange(var_di.n_dof, dtype=nm.int32)
 
+        self.n_dg_ebc = 0
+        self.dg_ebc_names = {}
+        self.dg_ebc = {}
+        self.dg_ebc_val = {}
+
+        self.n_dg_epbc = 0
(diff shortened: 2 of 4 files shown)


sfepy/base/conf.py
```diff
@@ -84,6 +85,9 @@ def transform_conditions(adict, prefix):
 
     return d2
 
+def transform_dgebcs(adict):
+    return transform_conditions(adict, "dgebcs")
```
rc, line 89: I prefer 'dgebc' for the prefix to be consistent.

sfepy/discrete/common/dof_info.py
```diff
@@ -7,9 +7,11 @@
 import numpy as nm
 import scipy.sparse as sp
 
-from sfepy.base.base import assert_, Struct, basestr
+from sfepy.base.base import assert_, Struct, basestr, output
```
rc, line 10: The debug output is not commented out - remove?

sfepy/discrete/common/dof_info.py
```diff
@@ -290,6 +292,19 @@ def _init_empty(self, field):
...
 
+        self.n_dg_ebc = 0
+        self.dg_ebc_names = {}
+        self.dg_ebc = {}
+        self.dg_ebc_val = {}
+
+        self.n_dg_epbc = 0
+        self.dg_epbc_names = []
+        self.dg_epbc = []
+
```
rc, line 303: Too many empty lines.

sfepy/discrete/common/dof_info.py
```diff
@@ -346,14 +361,21 @@ def map_equations(self, bcs, field, ts, functions, problem=None,
 
             active_bcs.add(bc.key)
 
-            if isinstance(bc, EssentialBC):
+            if isinstance(bc, DGEssentialBC):
```
rc, line 364: Just from the "new feature to the end" perspective, I would put the DG conditions after the FEM ones.

sfepy/discrete/common/dof_info.py
```diff
@@ -346,14 +361,21 @@ def map_equations(self, bcs, field, ts, functions, problem=None,
...
+            elif isinstance(bc, EssentialBC):
                 ntype = 'EBC'
                 region = bc.region
-
-            else:
+            elif isinstance(bc, PeriodicBC):
                 ntype = 'EPBC'
                 region = bc.regions[0]
 
+            # output("Treating {} {}".format(ntype, bc.name))
```
rc, line 377: remove?

sfepy/discrete/common/dof_info.py
```diff
@@ -460,6 +507,8 @@ def map_equations(self, bcs, field, ts, functions, problem=None,
         self.eq_ebc = nm.atleast_1d(ii.squeeze())
         self.val_ebc = nm.atleast_1d(val_ebc[ii].squeeze())
         self.master = nm.argwhere(master_slave > 0).squeeze()
+        # add axis in case we squeezed too hard
+        self.master = self.master if self.master.shape else self.master[None]
```
rc, line 511: When is this needed? Would using `nm.atleast_1d()` help as for `self.val_ebc`?

sfepy/discrete/conditions.py
```diff
@@ -29,6 +29,8 @@ def get_condition_value(val, functions, kind, name):
     elif (isinstance(val, Function) or nm.isscalar(val)
           or isinstance(val, nm.ndarray)):
         fun = val
+    elif isinstance(val, tuple):
```
rc, line 32: stylistic nitpicking: the other branches have an empty line above.

Would it make sense to allow lists as well? (See `is_sequence()` in `sfepy/base/base.py`).

sfepy/discrete/conditions.py
```diff
@@ -55,7 +57,16 @@ def from_conf(conf, regions):
         for key, cc in six.iteritems(conf):
             times = cc.get('times', None)
 
-            if 'ebc' in key:
+            if 'dgebc' in key:
```
rc, line 60: Just from the "new feature to the end" perspective, I would put the DG conditions after the FEM ones. That is, above 'lcbc'...

sfepy/discrete/problem.py
```diff
@@ -645,13 +645,21 @@ def set_bcs(self, ebcs=None, epbcs=None, lcbcs=None):
             conf_ebc = get_default(ebcs, self.conf.ebcs)
             self.ebcs = Conditions.from_conf(conf_ebc, self.domain.regions)
 
+            conf_dgebc = self.conf.get("dgebcs", {})
+            self.ebcs.extend(Conditions.from_conf(conf_dgebc,
+                                                  self.domain.regions))
+
```
rc, line 651: Is it possible to mix ebcs and dgebcs?

rc:
Thanks! Yes, smaller chunks are better ;)

sfepy/discrete/conditions.py
zitkat: Ok, should I put it before or after IC?

rc: I thought after epbc, before lcbc, but after ic would do as well.

sfepy/discrete/common/dof_info.py
zitkat: There is an issue with class hierarchy `DGEssentialBC` is subclass of `EssentialBC`, hence it is also instance of it.

sfepy/discrete/conditions.py
zitkat: I have encoutered issue similiar to the class hierarchy problem in dof_info, if the condition is "dgebc" the `key` contains "ebc", so they can't swapped without changing the decision logic to something like
```python
if (not "dgebc" in key) and ("ebc" in key):
   ...
```

sfepy/discrete/common/dof_info.py
rc: I see. Well, one could use `type(bc) == ...` but the cleanest way would be to make `ntype` and `region` (for periodic BCs) attributes of `bc` in the constructors in `conditions.py`. Then it would be more pythonic and the isinstance tests would not be needed. But in this PR we could just leave it as it is.

sfepy/discrete/conditions.py
rc: :) True, good catch. What about `key.startswith()`?

sfepy/discrete/problem.py
zitkat: No it is not, DGField does not allow setting facet DOFs, because technically there are none. It is however technically possible to mix epbc and dgepbc, although I'm not use if it makes sense.

sfepy/discrete/common/dof_info.py
zitkat: This is needed in 1D problems for which line
```python
self.n_epbc = self.master.shape[0]
```
bellow would fail. However since DG is using dgepbc for 1D problems it is not necessary.

rc:
Thanks for the updates!

Almost ready to merge, except one bug that manifested when running
```
python3 examples/diffusion/laplace_refine_interactive.py output
```
This code is not covered by regular tests (only by test_install.py), so it skipped detection. I proposed a fix in https://github.com/rc/sfepy/commits/dg-sfepy-bc-integration - try including the commit 1dee81f28ff682d9793f011e13c523af30664d22 whether it works with the DG code. If OK, push it here, and I will merge this PR.

zitkat:
The fix works fine with DG code. I have tried `test_install.py` and the `laplace_refine_interactive.py` passes but I had to edit the script to work under MS Windows, because it calls python as `python3`. I still get some failures, mostly due to missing packages but 
```python
out, err = check_output('%s ./simple.py examples/navier_stokes/stokes.py' % cmd)
eok += report(out, '...', -3, 5, '1.210678e-13', eps=1e-11)
```
fails due to high residuals.

rc:
It looks OK. The lower precision is probably related to a different (external) linear solver.

PR #600, opened 2020-04-15
fix saving 1D meshes in vtk and vtu formats

vlukes:


sfepy/discrete/fem/meshio.py
```diff
@@ -403,6 +403,11 @@ def write(self, filename, mesh, out=None, ts=None, **kwargs):
         if self.file_format in ['vtk', 'vtu']:
             ngkey = 'node_groups'
             cgkey = 'mat_id'
+            if coors.shape[1] != 3:
+                nnd, ndim = coors.shape
+                ndim = 3 - ndim
+                coors = nm.hstack([coors,
```
rc, line 409: meshio uses `numpy.pad()` - might be faster?

rc:
Thanks! LGTM, only one question regarding the padding function, see above.

sfepy/discrete/fem/meshio.py
vlukes: I hope that this way is fast enough. From my point of view, this solution is more readable than   using `numpy.pad()`.

rc: OK

PR #603, opened 2020-04-20
fix meshio read_data()

### LINKED ISSUE #601: MeshioLibIO.read_data returns only cell data for `gmsh:ref`
Hi, during my tinkering with I/O for gmsh  (https://github.com/sfepy/sfepy/pull/593) I encountered following issue: The `write` method of `MeshioLibIO` writes `node_groups` and `mat_id` as point and cell data, both with label `'%s:ref' % self.file_format`. When reading data using `read_data` the vertex data are then overwritten by cell data in `out` dictionary since both have the same labels used as keys.

vlukes:
This PR should fix #601.


rc:
LGTM, I will merge after @zitkat confirms fixing #601 .

PR #604, opened 2020-04-21
lower memory usage of tests

rc:
See also #595.


rc:
The black curve is with explicit gc:
memory_profiles_run_tests

vlukes:
Subjectively, I can observe about 1/3 reduction of peak memory usage of the tests. So it seems good!

PR #611, opened 2020-05-09
Force nm.int64 in nm.prod to prevent overflow on win64 platform

### LINKED ISSUE #610: Integer overflow in matrix size check win64
So I have been pushing the DG method a bit and got to huge matrices for implict problems and encountered an issue in Equations.create_graph() https://github.com/sfepy/sfepy/blob/02efa4595881910fc781daa8162d44e835c816a0/sfepy/discrete/equations.py#L466-L467
and 
https://github.com/sfepy/sfepy/blob/02efa4595881910fc781daa8162d44e835c816a0/sfepy/discrete/equations.py#L485-L486
When the matrix is large, say 65536 x 65536, which for DG corresponds to 64 x 64 mesh with 3rd order elements 16 DOFs each, which does not seem that excessively large to me, the int32 arithmetic used by numpy.prod overflows. This does not cause problems for smaller or larger (sic) matrices but 65536 x 65536 is exactly zero in 32-bit artithmetic and one gets `'no matrix (zero size)!'` message and subsequently error because the method does not return the matrix. 
Forcing int64 by 
```
nm.prod(shape, dtype=nm.int64)
```
would fix it.

zitkat:
Fixes #610

### DIFF
--- a/sfepy/discrete/equations.py
+++ b/sfepy/discrete/equations.py
@@ -463,7 +463,7 @@ class Equations(Container):
         shape = get_default(shape, self.variables.get_matrix_shape())
 
         output('matrix shape:', shape, verbose=verbose)
-        if nm.prod(shape) == 0:
+        if nm.prod(shape, dtype=nm.int64) == 0:
             output('no matrix (zero size)!')
             return None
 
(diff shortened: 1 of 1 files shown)


sfepy/discrete/equations.py
```diff
@@ -483,7 +483,7 @@ def create_matrix_graph(self, any_dof_conn=False, rdcs=None, cdcs=None,
 
         output('...done in %.2f s' % timer.stop(), verbose=verbose)
         output('matrix structural nonzeros: %d (%.2e%% fill)' \
-               % (nnz, float(nnz) / nm.prod(shape)), verbose=verbose)
+               % (nnz, float(nnz) / nm.prod(shape, dtype=nm.int64)), verbose=verbose)
```
rc, line 486: Too long line...

rc:
Thanks! What about defining `size = nm.prod(shape, dtype=nm.int64)` and then using that?

rc:
LGTM, thanks! Only the last commit name/e-mail is strange - shall I fix it before merge?

zitkat:
Oh, I did the change from github and did not realize it will be this messy. You can fix it 
or squash the commits, it's up to you.

rc:
Merged after fixing the names.

PR #612, opened 2020-05-17
Discontinous Galerkin method

zitkat:
Hi, I have edited code for DG into form I'm willing to present for PR. I have tried to reduce the amount of changes but there is still a lot to unpack:

* Bulk of the method is in `sfepy/discrete/dg`:
  * `fields.py`
  * `limiters.py`  Limiters are kind of new to sfepy, they basically amount to poststep hooks for time stepping solvers, however they require connectivy information, hence they recieve field of the limited variable upon initialization. This I think would deserve more general treatment, we could exclude them from PR and resolve this later.
  * `dg_poly_spaces.py`: `Legendre*PolySpce` classes would fit nicely into `sfepy/discrete/fem/poly_spaces.py` as it would enable creating legendre discrete polyspace through `PolySpace.any_from_args`. However I did not want to mix classical  FEM and DG code. Note that `Legendre*PolySpace` can't be simply imported due to cyclic import. Maybe mechanism similiar to field registration would help here?
  * `utils` contain functions used by examples and utilities for plotting 1D DG results, these could be moved to scripts?

* Other components are in corresponding folders:
  * `terms_dg.py`: contains all DG specific terms implemented using einsums, there is definetly a lot of improvements that can be done, I welcome any suggestions.
   * `ts_dg_solvers.py`: contains explicit time stepping solvers used in examples, they are implemented a bit naively, but they work. They also contain mechanism for employing limiters which might need revision,
  * `examples/dg`: I have selected 5 examples runnable with `simple.py` and one imperative for testing. I have many more, we can later select those most ilustrative.
  * documentation: I have added `.rst` files pointing to DG source files and added reference to them to `developer_guide`.

* Changes to other files include:
  * all changes registering or creating new DG specific classes
  * gmshio: this PR is essentially blocked by #593 I have included changes to meshio.py for 
   convenience, but merging will be possible only after resolving #593 and rebasing.
   * I have also included some minor convinience changes in:
	`sfepy/mesh/mesh_generators.py`, 
	`sfepy/base/plotutils.py`.

I have included several TODOs with questions to bring attention to parts I would welcome some advice on or I feel need revision by you.


zitkat:
I only now noticed that `tests/test_term_call_modes.py` fail for DG terms, this is beacause DG terms raise `ValueError` when someone tries to use them with field other than DGField.

tests/test_term_call_modes.py
```diff
@@ -196,7 +196,9 @@ def test_term_call_modes(self):
                 domain_geometry = '%d_%s' % (domain.shape.dim, domain_geometry)
 
             for _, term_cls in ordered_iteritems(term_table):
-                if not domain_geometry in term_cls.geometries:
+                if (domain_geometry not in term_cls.geometries) \
+                   or ("dg" in term_cls.name) \
+                   or (term_cls.name == "dw_ns_dot_grad_s"):
```
zitkat, line 201: > I only now noticed that `tests/test_term_call_modes.py` fail for DG terms, this is beacause DG terms raise `ValueError` when someone tries to use them with field other than DGField.

I have looked into it and there are two issues

- one fundamental: the DG terms require boundary condition data. I don't think adding those to `test_term_call_modes.py` is a good idea so I omited DG terms  and will provide my own tests

- second a bit messy: `dw_ns_dot_grad_s`  implemented in terms_dg.py ([here](https://github.com/sfepy/sfepy/blob/6f2412096ee1ec1cfb73e7f29861e6a5706ac677/sfepy/terms/terms_dg.py#L677)) requires as parameters callables which the test can't provide. I dug a bit into it and do not see clean way to patch it, so I omit it from the test and will provide my own test too.

rc:
Hi @zitkat,

great work, all tests pass! I am going through the code as time permits...

Notes so far:
- Use of f-strings means Python >= 3.6, but it is (right?) the only feature so far? Probably OK.
- CFL or cfl as argument name? Or parCFL?
- Convert docstrings to numpydoc standard (also Params -> Parameters in numpydoc section names)
- [later] `local_register_function()` -> `register_function()`? It might be handy in `sfepy.dicrete.functions`, add also `make_register_function(functions)` to avoid using global `functions` dict.
- Examples: it would be nice if the docstrings followed the scheme of other examples (e.g. `examples/diffusion/time_advection_diffusion.py`). The first paragraph is displayed in the [gallery](http://sfepy.org/doc-devel/examples/gallery.html).
- Docs:
  - add DG terms to the term table (see `script/gen_term_table.py`)
  - add DG examples to gallery and docs.
- [later] commented-out variants in examples (e.g. in `imperative_dg_burgess_1D.py`) - or how to make examples configurable:
  - imperative: turn them into proper scripts with `parse_args()`, `main()`...
  - declarative: (this is done, but for completenes) use `define()` function
- Too long lines: I provided some formatting examples in  the notes to `imperative_dg_burgess_1D.py` above. BTW. isn't it Burgers instead of Burgess?

rc:
More notes:

- Use `test_dg_*()` naming scheme for the new test files?
- Many files have unused imports (pyflakes reports those).
- PolySpaces: to get rid of the cyclic import problem, we could
  - either put the generic functions to `sfepy/discrete/common/poly_spaces.py`, and use something similar as in `Field.from_conf()` - or we could just import the dg, fem modules inside the `PolySpace.any_from_args()`.
  - or simply just put all the classes into a single file into `sfepy/discrete/common/poly_spaces.py`?
  - I can do the refactoring, if that helps.

rc:
Thanks for all the updates, let us merge it. Issues are likely in such a big PR, but those can be resolved separately one by one.

PR #635, opened 2020-08-24
Mumps5.2

vlukes:


rc:
Thanks! LGTM, just is the `from __future__ import absolute_import` needed, when sfepy is now Python 3 only?

vlukes:
Using `from __future__ import absolute_import` suppresses some Sphinx warnings when generating HTML.

rc:
OK!

PR #639, opened 2020-09-07
Add generalized Yeoh term

heczis:
Naive implementation in python + an example.
The term may be used to model highly filled rubbers.

### DIFF
--- a/sfepy/terms/terms_hyperelastic_tl.py
+++ b/sfepy/terms/terms_hyperelastic_tl.py
@@ -52,6 +52,134 @@ class NeoHookeanTLTerm(HyperElasticTLBase):
     stress_function = staticmethod(terms.dq_tl_he_stress_neohook)
     tan_mod_function = staticmethod(terms.dq_tl_he_tan_mod_neohook)
 
+class GenYeohTLTerm(HyperElasticTLBase):
+    r"""
+    Hyperelastic generalized Yeoh term [1]. Effective stress
+    :math:`S_{ij} = 2 p K (I_1 - 3)^{p-1} J^{-\frac{2}{3}}(\delta{ij} -
+    \frac{1}{3}C_{kk}C{ij}^{-1})`.
+
+    :Definition:
+
+    .. math::
+        \int_{\Omega} S_{ij}(\ul{u}) \delta E_{ij}(\ul{u};\ul{v})
+
+    :Arguments:
+        - material : :math:`p, K`
+        - virtual  : :math:`\ul{v}`
+        - state    : :math:`\ul{u}`
+
+    [1] Travis W. Hohenberger, Richard J. Windslow, Nicola M. Pugno,
+    James J. C. Busfield. Aconstitutive Model For Both Lowand High
+    Strain Nonlinearities In Highly Filled Elastomers And
+    Implementation With User-Defined Material Subroutines In
+    Abaqus. Rubber Chemistry And Technology, Vol. 92, No. 4,
+    Pp. 653-686 (2019)
+    """
+    name = 'dw_tl_he_genyeoh'
+    family_data_names = ['det_f', 'tr_c', 'sym_inv_c']
+    arg_shapes = {'material' : '1, 2',
+                  'virtual' : ('D', 'state'), 'state' : 'D'}
+    geometries = ['3_4', '3_8']
+
+    @staticmethod
+    def _get_single_stress(i_1, i_3, c_inv, coef, exp):
+        _bracket = (i_3**(-1 / 3) * i_1 - 3.)
+        _bracket_pow = _bracket**(exp - 1) \
+            if (_bracket > 0. or exp >= 1.) else 1.
+        out = 2 * exp * coef * _bracket_pow \
+            * i_3**(-1 / 3) * (nm.eye(3) - i_1 * c_inv / 3)
+        return out
+
+    def stress_function(self, out, mat, *fargs, **kwargs):
+        coef, exp = mat[:, :, :, :1], mat[:, :, :, 1:]
+        det_f, tr_c, inv_c = fargs
+
+        i_3 = det_f**2
+        ident = nm.array([1., 1, 1, 0, 0, 0])
+
+        n_cells, n_qps, _, _ = out.shape
+        for cell in range(n_cells):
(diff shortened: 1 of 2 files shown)


sfepy/terms/terms_hyperelastic_tl.py
```diff
@@ -52,6 +52,133 @@ class NeoHookeanTLTerm(HyperElasticTLBase):
...
+    James J. C. Busfield. Aconstitutive Model For Both Lowand High
+    Strain Nonlinearities In Highly Filled Elastomers And
+    Implementation With User-Defined Material Subroutines In
+    Abaqus. Rubber Chemistry And Technology, Vol. 92, No. 4,
+    Pp. 653-686 (2019)
+    """
+    name = 'dw_tl_he_genyeoh'
+    family_data_names = ['det_f', 'tr_c', 'sym_inv_c']
+    arg_shapes = {'material' : '1, 2',
+                  'virtual' : ('D', 'state'), 'state' : 'D'}
```
rc, line 81: Add
```
    geometries = ['3_4', '3_8']
```
here, as the term work only in 3D. This should solve the CI failures.

examples/large_deformation/hyperelastic_gen_yeoh_up.py
```diff
@@ -0,0 +1,399 @@
...
+
+    ax_difference.plot(stretch, stress_fem - stress_analytic, '.-')
+
+    ax_stress.legend(loc='best').set_draggable(True)
+    ax_stress.set_ylabel(r'nominal stress $\mathrm{[Pa]}$')
+    ax_stress.grid()
+
+    ax_difference.set_ylabel(r'difference in nominal stress $\mathrm{[Pa]}$')
+    ax_difference.set_xlabel(r'stretch $\mathrm{[-]}$')
+    ax_difference.grid()
```
rc, line 193: Use `plt.tight_layout()` to make all labels visible.

rc:
Thanks! I will try to look at the PR tomorrow, for now I have just the two comments above.

examples/large_deformation/hyperelastic_gen_yeoh_up.py
```diff
@@ -0,0 +1,399 @@
...
+.. math::
+    W =
+      K_1 \, \left( \overline I_1 - 3 \right)^{m}
+      +K_2 \, \left( \overline I_1 - 3 \right)^{p}
+      +K_3 \, \left( \overline I_1 - 3 \right)^{q}
+
+where :math:`\overline I_1` is the first main invariant of the deviatoric part
+of the right Cauchy-Green deformation tensor :math:`\ull{C}`. The coefficients
+:math:`K_1, K_2, K_3` and exponents :math:`m, p, q` are material parameters.
+Only the first term is used in this example for simplicity.
```
rc, line 16: Could you explain better the correspondence of the above parameters to the ones used in the term docstring (p, K == m, K_1?).

So the general model can be obtained by adding three dw_tl_he_genyeoh terms together?

examples/large_deformation/hyperelastic_gen_yeoh_up.py
```diff
@@ -0,0 +1,399 @@
...
+
+The three-term generalized Yeoh model is meant to be used for modelling of
+filled rubbers. The following choice of parameters is suggested [1] based on
+experimental data and stability considerations:
+
+  :math:`K_1 > 0`
+  :math:`K_2 < 0`
+  :math:`K_3 > 0`
+  :math:`0.7 < m < 1`
+  :math:`m < p < q`
```
rc, line 70: This renders to a single line, so use commas and a dot after the expressions.

examples/large_deformation/hyperelastic_gen_yeoh_up.py
```diff
@@ -0,0 +1,399 @@
...
+
+Usage Examples
+--------------
+
+Default options::
+
+  $ python examples/large_deformation/hyperelastic_tl_up_interactive.py
```
rc, line 77: Here and below, use the correct example name ;)

rc:
Hi again. I have added more comments and have some suggestions:

1. It would be nice to have explained in the example docstring how to visualize the results with resview.py and/or postproc.py to see, for example, the stress on a deformed mesh.

    I have tried `python resview.py --fields=p:f1:wu:p1 domain.??.vtk` and it works, but having a possibility to fix the view between time steps would be nice (pinging @vlukes).

    With postproc.py: `python postproc.py -b --wire --only-names=u -d 'u,plot_displacements,rel_scaling=1,color_kind="scalars",color_name="p"' --view=6.06e+01,4.35e+01,5.19e+00,9.01e-01,-8.31e-01,-5.02e-01 --roll=3.79e+00 domain.??.vtk`. Using tensors for the colors is broken for .vtk, it works for .h5, but only for meshes with more than 1 element (probably an array squeezing/reshaping problem).

2. The interactive example file names should end with _interactive.py - what about naming it `gen_yeoh_tl_up_interactive.py`?

vlukes:
Hi @rc,

what exactly do you mean by "a possibility to fix the view between time steps"?

rc:
> Hi @rc,
> 
> what exactly do you mean by "a possibility to fix the view between time steps"?

If you try the example, press 'r' on the final time step to see the whole mesh and then go back in time, the mesh 'moves' (even the point that is fixed by EBCs), because probably the camera focus moves.

vlukes:
OK, I see. But this strange behavior is reproducible only for this data. All the other examples I have tested are fine. Maybe too big change in geometry between time steps.

rc:
Maybe, anyway, for a quick visualization the current way is OK.

heczis:
All suggestions should be resolved now. I somewhat rewrote the first paragraph of the example - does it make the model understandable?

rc:
Thanks, yes, now it is clear. LGTM!

I wonder if having the long example lines as single lines without the continuation character `\` was better - it would definitely make copy-paste-editing those lines to a terminal easier. What do you think?

Would you like to add the resview.py postprocessing example I mentioned above, or I merge now?

heczis:
I added resview.py examples. Also checked the results again - should be fine.

rc:
OK, thanks!

PR #640, opened 2020-09-10
fix meshio interface to accept ascii/binary file format variants

vlukes:
Now, ascii/binary file format variants can be used in `file_format` option:

```
options = {
    'file_format': 'vtk-ascii',
}
```


rc:
Thanks!

PR #664, opened 2021-01-11
update CoefNN: allow different dimensions for row and column

vlukes:
This calculates a coefficient of dimensions 2x3 (for 3D problem):
```
  'dim': (dim - 1, dim),
  'class': cb.CoefNN,
```

### DIFF
--- a/sfepy/homogenization/coefs_base.py
+++ b/sfepy/homogenization/coefs_base.py
@@ -855,7 +855,7 @@ class VolumeFractions(MiniAppBase):
         return vf
 
 
-class CoefNN(MiniAppBase):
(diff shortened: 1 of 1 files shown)


rc:
Thanks! Wouldn't it be more logical to name the coefficient then CoefNM? Reusing CoefNN seems a bit misleading.

vlukes:
> Thanks! Wouldn't it be more logical to name the coefficient then CoefNM? Reusing CoefNN seems a bit misleading.

By default, the term is dim*dim. This PR is just an extension of the current functionality. But you are right, CoefMN makes more sense.

rc:
Thanks, LGTM!

PR #665, opened 2021-01-11
fix volumes in DensityVolumeInfo class

vlukes:
I think that all the coefficients should be divided by the whole volume of the reference cell, i.e. by `true_volume`.

### DIFF
--- a/sfepy/homogenization/coefs_phononic.py
+++ b/sfepy/homogenization/coefs_phononic.py
@@ -718,7 +718,7 @@ class DensityVolumeInfo(MiniAppBase):
 
         return Struct(name='density_volume_info',
                       average_density=average_density,
-                      total_volume=total_volume,
+                      total_volume=true_volume,
                       volumes=volumes,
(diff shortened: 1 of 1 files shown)


rc:
Let me recall: total_volume is the volume of the mesh, i.e. it does not count holes?

vlukes:
> Let me recall: total_volume is the volume of the mesh, i.e. it does not count holes?

Yes, `total_volume` does not count holes.

rc:
OK, thanks, LGTM!

rc:
Am I right to assume that the homogenized density should be with the holes included? It makes sense, otherwise single density material with holes would have the same density as without.

vlukes:
> Am I right to assume that the homogenized density should be with the holes included? It makes sense, otherwise single density material with holes would have the same density as without.

Yes, you are right. `average_density` presents the homogenized density in the subsequent calculations.

rc:
Thanks!

PR #678, opened 2021-03-29
Remove med mesh files

rc:
Trying to fix CI failures. The med files are not valid anyway (cannot be converted using current `script/convert_mesh.py`).


rc:
FWIW, I have fixed the CI failure in #776 by using conda-forge channel conda packages. But meshio > 4.3.8 is not compatible with the  removed .med files, so the failure would re-appear in future. @vlukes, what about removing the meshes that are not used anywhere and covered by meshio?

PR #681, opened 2021-04-13
new example application

vlukes:
New example application: Two-scale numerical simulation of a large deforming fluid-saturated porous structure


rc:
Thanks, merging!

Would it be possible to add more results figures (such as what happens on the microscale) to the example page?

vlukes:
Yes, I would like to add some more results to the example page. This is the first quick shot.

PR #682, opened 2021-04-22
New sd term

vlukes:
- new piezo coupling sensitivity term: `d_sd_piezo_coupling`
- new test of sensitivity terms (except `d_sd_convect`, because `dw_convect` can not be evaluated)

### DIFF
--- a/sfepy/terms/terms_multilinear.py
+++ b/sfepy/terms/terms_multilinear.py
@@ -209,9 +209,11 @@ class ExpressionArg(Struct):
             obj = ExpressionArg(name='_'.join(term.arg_names[ii]), arg=arg,
                                 kind='ndarray')
 
+        elif isinstance(arg, tuple) and isinstance(arg[0], nm.ndarray):
+            obj = ExpressionArg(name=arg[1], arg=arg[0], kind='ndarray')
+
         else:
-            raise ValueError('unknown argument type! ({})'
-                             .format(type(arg)))
+            raise ValueError('unknown argument type! ({})'.format(type(arg)))
 
         return obj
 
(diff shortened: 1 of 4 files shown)


sfepy/terms/terms_multilinear.py
```diff
@@ -746,8 +740,8 @@ def make_function(self, texpr, *args, diff_var=None):
 
         if einfo.eargs is None:
             einfo.eargs = [
-                ExpressionArg.from_term_arg(arg, self, self.expr_cache)
-                for arg in args
+                ExpressionArg.from_term_arg(arg, arg_name, self, self.expr_cache)
+                for arg, arg_name in zip(args, self.arg_types)
```
rc, line 744: This does not work, because the `make_function()` arguments can be (and are) different from the term arguments.

vlukes: Yes, you are right. My intention is to avoid the second arguments check in `make_function()`. The new version should work. The tests pass.

Now, the names of ndarray `args` can be specified as:  

```
self.make_function('ij,i,j', (mat1, 'material_1'), (mat2, 'material_2'), ...)
```

sfepy/terms/terms_multilinear.py
```diff
@@ -745,9 +744,11 @@ def make_function(self, texpr, *args, diff_var=None):
             return einfo.eval_einsum
 
         if einfo.eargs is None:
+            equl_args = len(args) == len(self.ats)
```
rc, line 747: This does not check that args really equal term.args ;)

What about just passing `arg_pos=k` always, and setting the name to `'par{}'.format(arg_pos)`?

rc:
Thanks! Did the greedy optimization help?

sfepy/terms/terms_multilinear.py
vlukes: If `len(args)` is equal to `len(term.args)` then `arg_pos=k` may be reasonable. If not, the argument name shoud be specified by a tuple `(arg, arg_name)`. Otherwise, the name is set to `unknown_parameter`.

sfepy/terms/terms_multilinear.py
vlukes: `arg_pos` argument removed. All "non-standard" arguments will be treated with tuples `(arg, arg_name)`.

rc:
LGTM, merging!

PR #683, opened 2021-04-27
Make de convect evaluable

vlukes:


rc:
Thanks, pretty easy, right? :) Only add also the integral with the parameter variables to the docstring, like e.g. in de_laplace.and other terms that can be evaluated.

vlukes:
> Thanks, pretty easy, right? :) Only add also the integral with the parameter variables to the docstring, like e.g. in de_laplace.and other terms that can be evaluated.

Yes, really easy! The term docstring is updated.

rc:
Thanks, merging!

PR #684, opened 2021-05-05
New SDLinearTractionTerm

vlukes:
- fix `geme_mulAVSB3py()` - values were calculated for one cell only
- new `SDLinearTractionTerm` - calculates shape sensitivity of  `LinearTractionTerm`
- new test for `SDLinearTractionTerm`

### DIFF
--- a/sfepy/terms/terms_surface.py
+++ b/sfepy/terms/terms_surface.py
@@ -116,12 +116,13 @@ class SDLinearTractionTerm(Term):
         - material  : :math:`\ull{\sigma}`
         - parameter : :math:`\ul{v}`
     """
+
     name = 'd_sd_surface_ltr'
     arg_types = ('opt_material', 'parameter', 'parameter_mv')
     arg_shapes = [{'opt_material': 'S, 1', 'parameter': 'D',
-                   'parameter_mv': 'D'}, {'opt_material' : '1, 1'},
-                  {'opt_material' : 'D, 1'}, {'opt_material' : 'D, D'},
-                  {'opt_material' : None}]
+                   'parameter_mv': 'D'}, {'opt_material': '1, 1'},
+                  {'opt_material': 'D, 1'}, {'opt_material': 'D, D'},
+                  {'opt_material': None}]
     integration = 'surface'
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


sfepy/terms/terms_surface.py
```diff
@@ -18,9 +19,10 @@ class LinearTractionTerm(Term):
     :math:`\ul{f}` for a traction vector, and itself for a stress tensor.
 
     The material parameter can have one of the following shapes: 1 or (1, 1),
-    (D, 1), (S, 1). The symmetric tensor storage is used in the last case: in
-    3D S = 6 and the indices ordered as :math:`[11, 22, 33, 12, 13, 23]`, in 2D
-    S = 3 and the indices ordered as :math:`[11, 22, 12]`.
+    (D, 1), (S, 1), or (D, D) in the `eval` mode. The symmetric tensor storage
+    is used in the last case: in 3D S = 6 and the indices ordered as
```
rc, line 23: Make (S, 1) last so that it is the last case.

sfepy/terms/terms_surface.py
```diff
@@ -95,6 +101,86 @@ def set_arg_types( self ):
...
+                  {'opt_material': None}]
+    integration = 'surface'
+
+    @staticmethod
+    def d_fun(out, traction, val, grad_mv, div_mv, sg):
+
+        tdim, tdim2 = (None, None) if traction is None else traction.shape[2:]
+        dim = val.shape[2]
+        sym = (dim + 1) * dim // 2
+        nel, nqp = div_mv.shape[:2]
```
rc, line 134: In `get_eval_shape()` there is `n_el`, `n_qp`...

rc:
Thanks! It seems nobody used the tensor mode ;)

sfepy/terms/terms_surface.py
vlukes: `(D, D)` is possible in  the `eval` mode only, so I modified the docstring in this way.

sfepy/terms/terms_surface.py
vlukes: `SDLinearTractionTerm.d_fun()` is the static method so I would have to call `div_mv.get_data_shape()`, but it requires `integrals`, etc. I think the current solution is simple and reliable.

rc: Yes, just that you use different names for the same things (nel x n_el, ...)

sfepy/terms/terms_surface.py
rc: OK, so what about
```
The symmetric tensor storage (S, 1) is as follows: in 3D S = 6 and the indices ordered as
```

vlukes: Yes, I see. I did not noticed "in the last case".

sfepy/terms/terms_surface.py
vlukes: renamed: `nel`, `nqp` --> `n_el`, `n_qp`

rc:
LGTM, thanks!

PR #695, opened 2021-06-04
link example application in index.rst

rc:
@vlukes, what do you think about the proposed changes?


vlukes:
Good idea. It can improve  the visibility of examples.

rc:
OK, will merge it then.

PR #699, opened 2021-06-23
Material shape update

vlukes:
The PR allows for material shape (1, M, N). It is tiled into (0, 8, 1, 1) which is which is handled by `FMFSetCelX1()`. This can save a lot of memory.

### DIFF
--- a/doc/users_guide.rst
+++ b/doc/users_guide.rst
@@ -1055,6 +1055,10 @@ The function needs to be registered in `Functions`_::
       'get_pars' : (get_pars,),
   }
 
+If a material parameter has the same value in all quadrature points, than it is
+not necessary to repeat the constant and the array can be with shape
+`(1, n_row, n_col)`.
+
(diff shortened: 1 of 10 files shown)


rc:
Thanks! Could you modify an example or add a test so that the new code path is covered in testing?

vlukes:
> Thanks! Could you modify an example or add a test so that the new code path is covered in testing?

Yes, I have modified the following examples:

- linear_elasticity/linear_viscoelastic.py
- multi_physics/biot_short_syntax.py
- multi_physics/piezo_elasticity.py

rc:
Thank you, LGTM!

Or maybe add a note to https://sfepy.org/doc-devel/users_guide.html#defining-material-parameters-by-functions - it seems like an important feature for users, so let's have it documented.

vlukes:
OK, user's guide is updated.

rc (approved):

PR #700, opened 2021-06-25
Constant materials

vlukes:
Material parameters with constant value are now created with shape (1, nqp, n, m).

### DIFF
--- a/sfepy/discrete/common/extmods/fmfield.h
+++ b/sfepy/discrete/common/extmods/fmfield.h
@@ -42,6 +42,8 @@ typedef struct FMField {
 #define FMF_PtrFirst( obj ) ((obj)->val0)
 #define FMF_PtrCurrent( obj ) ((obj)->val)
 #define FMF_PtrCell( obj, n ) ((obj)->val0 + (n) * (obj)->cellSize)
+#define FMF_PtrCellX1( obj, n ) ((obj)->nCell > 1 \
+  ? (obj)->val0 + (n) * (obj)->cellSize : (obj)->val0)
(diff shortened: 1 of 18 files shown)


rc:
Thanks!

I have just a few remarks:
- I propose renaming `mat_constant` argument of `ConstantFunction` to `ret_constant`, or `no_tile`.  Functions are not used only for evaluating materials.
- Not sure about placement of `make_full_mat_array()` - what about making it a static method of Term? And/or rename it to `tile_mat()`?
- `ETermBase` needs also full sizes of materials -> `ExpressionArg` should do it.

vlukes:
> Thanks!
> 
> I have just a few remarks:
> 
>     * I propose renaming `mat_constant` argument of `ConstantFunction` to `ret_constant`, or `no_tile`.  Functions are not used only for evaluating materials.
>

OK, argument renamed to `no_tile`.
 
>     * Not sure about placement of `make_full_mat_array()` - what about making it a static method of Term? And/or rename it to `tile_mat()`?
> 

I have moved the function into `Term` class as `tile_mat()` static method.

>     * `ETermBase` needs also full sizes of materials -> `ExpressionArg` should do it.

Is it really needed? Are there tests for the multilinear terms?

vlukes:
> Previously, the tests/test_term_call_modes.py test failed on einsums, now it passes. Actually einsum should support bradcasting, so it may work?

Yes, einsum support broadcasting so I think it should work.

rc:
I do not see `no_tile` in the sources?

vlukes:
> I do not see `no_tile` in the sources?

OK, commit is now on the right branch.

rc (approved):

PR #702, opened 2021-06-28
Constant materials small fixes

rc:
PhysicalQPs fix is needed when a higher order approximation is used (n_qp is large).


vlukes (approved):
:+1:

PR #716, opened 2021-07-29
new iga-interactive example

livoire13:
Sfepy does not feature any example for isogeometric analysis in interactive mode. I hereby propose to add a simple IGA-interactive example for solving Laplace equation on a two-dimensional domain. The underlying NURBS-patch is created by means of igakit CAD built-in functions and then saved to the sfepy custom '.iga' file format.

I believe this example can be helpful for sfepy users interested in iga-domain creation and who favor the interactive mode over the 'problem description file' way (which is already featured in three examples so far).


examples/diffusion/laplace_iga_interactive.py
```diff
@@ -0,0 +1,202 @@
...
+NURBS domain using the isogeometric analysis (IGA) approach, using commands
+for interactive use.
+
+This script allows the creation of a customisable NURBS surface using igakit
+built-in CAD routines, which is then saved in custom HDF5-based files with
+.iga extension.
+
+View the results using::
+
+  $ ./postproc.py concentric_circles.vtk
```
rc, line 13: Use
```
  python3 postproc.py concentric_circles.vtk
```
here,

examples/diffusion/laplace_iga_interactive.py
```diff
@@ -0,0 +1,202 @@
...
+"""
```
rc, line 14: Please add:
```
import sys
sys.path.append('.')
```
- it is a hack that allows running the example from the source code directory on platforms, where the current directory (cwd) is not automatically added to sys.path (e.g. on a ecent kubuntu).

rc:
Thanks @livoire13 for a nice example, it works fine! I have just a few comments:

- Currently the code would be executed also when importing the example - put, please, the code into a function called main(), as in e.g. examples/diffusion/laplace_refine_interactive.py.
- There are several options (patch shape, viewpatch...) to tweak - would you consider adding command line options for those? Again, you can follow the laplace_refine_interactive.py example.
- See also the comments above in the code.
- There is some trailing whitespace - can you setup your editor/IDE to remove it automatically?

livoire13:
Thanks for the review @rc, I am going to take care of these various required edits.

examples/diffusion/laplace_iga_interactive.py
```diff
@@ -8,13 +8,31 @@
 built-in CAD routines, which is then saved in custom HDF5-based files with
 .iga extension.
 
+Usage Examples
+--------------
+
+Default options, storing results in this file's parent directory::
+
+  $ python3 laplace_iga_interactive.py
```
rc, line 16: I missed this: the other examples use the path relative to the main sfepy directory, i.e. use `examples/diffusion/laplace_iga_interactive.py` here.

examples/diffusion/laplace_iga_interactive.py
```diff
@@ -136,67 +154,113 @@ def insertUniformly(U,n):
...
+    filename_domain = data_dir + '/meshes/iga/concentric_circles.iga'
+    domain = IGDomain.from_file(filename_domain)
 
-# Setting a Domain instance
-filename_domain = data_dir + '/meshes/iga/concentric_circles.iga'
-domain = IGDomain.from_file(filename_domain)
+    # Sub-domains
+    omega = domain.create_region('Omega', 'all')
+    Gamma_out = domain.create_region('Gamma_out', 'vertices of set xi01',
+                                    kind='facet')
```
rc, line 213: One more space needed...

rc:
Thanks, almost ready - I found two details, see the code comments.

When running
```
python3 examples/diffusion/laplace_iga_interactive.py --R1=0.7 --C2=0.1,0.1 --viewpatch
```
the results have a discontinuity:
snapshot
-- is it expected/correct?

examples/diffusion/laplace_iga_interactive.py
```diff
@@ -0,0 +1,266 @@
...
+            iplt.figure()
+            iplt.plot(srf)
+            iplt.show()
+        except ImportError:
+            iplt.use('matplotlib')
+            iplt.figure()
+            iplt.plot(srf)
+            iplt.show()
+
+def insertUniformly(U,n):
```
rc, line 118: `insert_uniformly` ...

livoire13:
Sorry for all these oversights! I am going to fix them soon.
I left an error line 219 "order = 3" instead of using the parser.

The discontinuity you exhibit is not expected at all. After delving into it, I found where it comes from:

`    field = Field.from_args('fu', nm.float64, 'scalar', omega,
                            approx_order='iga+{:d}'.format(order_increase),
                            space='H1', poly_space_base='iga')`

approx_order='iga+{:d}.format{order_increase}' seems to be the origin of the failure whenever order_increase > 0. This is weird because according to [https://sfepy.org/doc-devel/users_guide.html#id16](url), it is possible to adjust the order of the IGA field on the go (i.e. even when domain.nurbs.degrees are strictly smaller). Any thoughts on that?

If not rectifiable, I will force optional argument order=order when calling the `create_patch` function.

rc:
Thanks for the updates!
 
I can still see the discontinuity, but IMO it is due to the fact that the patch is not connected and the inner and outer circles start at different angles (the disconnected boundary is not perpendicular to the circles due to the excentricity). I have tried adding `srf.unclamp(1)` according to [1], but that does not work (sfepy cannot probably deal with such a patch). Can you try adding periodic BCs to regions given by xi10 and xi11 (see tests/test_conditions.py, around line 162)?

BTW. the `order_increase` variable is currently ignored in the example, so if the periodic BCs do not help, it can be removed.

[1] https://petiga-igakit.readthedocs.io/en/latest/TUTORIAL3.html

livoire13:
Hum, I believe you are right, this is quite annoying indeed...

I am not familiar with periodic BCs in Sfepy, so I haven't tried your suggestion yet. However, I came up with a quick fix in the creation of the patch itself. When creating NURBS-curves for circle, one can specify a start angle and an end angle. In order to make sure that the boundary is indeed perpendicular to the line connecting the two circles, here's what I do:

```# Geometry Creation
centers_direction = [C2[0]-C1[0], C2[1]-C1[1]]
if centers_direction[0]==0 and centers_direction[1]==0:
	start_angle = 0.0
else:
	start_angle = nm.arctan2(centers_direction[1], centers_direction[0])
c1 = circle(radius=R1, center=C1, angle=(start_angle, start_angle+ 2*pi))
c2 = circle(radius=R2, center=C2, angle=(start_angle, start_angle+ 2*pi))```

If I am not mistaken, the discontinuity issue disappears.

rc:
Yes, the results look continuous now. I forgot that IGA fields do work with periodic BCs anyway :)

However, in order not to mislead users, could you mention this problem and the workaround in the example docstring?

Otherwise LGTM, thanks for a nice example!

livoire13:
Sure, I'll do that!
Thanks for helping me cleaning up the code.

Cheers

rc:
Great, thanks! LGTM.

PR #726, opened 2021-09-23
resview.py: automatic factor determination

rc:
Glyphs/warp factor can be determined automatically as  percentage of the minimum bounding box size.

### DIFF
--- a/resview.py
+++ b/resview.py
@@ -278,10 +278,10 @@ def pv_plot(filenames, options, plotter=None, step=None,
         warp = opts.get('w', options.warp)  # warp mesh
         factor = opts.get('f', options.factor)
         if isinstance(factor, tuple):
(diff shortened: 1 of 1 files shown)


resview.py
```diff
@@ -273,6 +275,12 @@ def pv_plot(filenames, options, plotter=None, step=None,
 
         warp = opts.get('w', options.warp)  # warp mesh
         factor = opts.get('f', options.factor)
+        if isinstance(factor, tuple):
+            bnds = pipe[-1].bounds
+            ws = nm.diff(nm.reshape(pipe[-1].bounds, (-1, 2)), axis=1)
```
vlukes, line 282: Use `bnds` defined above.

rc:
Thanks! I actually had a bug there (testing it on unit data...). It should be OK now.

PR #734, opened 2021-10-25
resview.py:  support sfepy h5 no steps

rc:
Hi @vlukes, could you try it?

### DIFF
--- a/resview.py
+++ b/resview.py
@@ -134,16 +134,24 @@ def read_mesh(filenames, step=None, print_info=True, ret_n_steps=False):
         cache['n_steps'] = len(filenames)
     elif ext in ['.xdmf', '.xdmf3']:
         import meshio
-        from meshio._common import meshio_to_vtk_type
+        try:
+            from meshio._common import meshio_to_vtk_type
+
+        except ImportError:
+            from meshio._vtk_common import meshio_to_vtk_type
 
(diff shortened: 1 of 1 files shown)


rc:
Ping @vlukes :)

resview.py
```diff
@@ -116,6 +116,12 @@ def make_cells_from_conn(conns, convert_to_vtk_type):
 
     return cells, cell_type, offset
 
+def add_mat_id_to_grid(grid, mesh):
```
vlukes, line 119: Calling the function with `mesh` is unnecessary, `cell_groups` array would be sufficient.

rc:
Thanks! I have one more thing: the default position vector is `[0, 0, 1.6]` which does not work well for 2D meshes (z = 0, all plots are in one place...) Maybe put the nonzero offset to the last nonzero axis, based on the mesh bounding box?

vlukes:
> Maybe put the nonzero offset to the last nonzero axis, based on the mesh bounding box?

Yes, it may work.

rc:
> > Maybe put the nonzero offset to the last nonzero axis, based on the mesh bounding box?
> 
> Yes, it may work.

OK, have a look. Works for me :)

rc:
Let us merge this, possible tweaks can go to another PR.

PR #737, opened 2021-10-26
Tried to improve documentation of Material constructor. 

burnpanck:
Made `values=` and `**kwargs` completely equivalent instead of almost equivalent

### DIFF
--- a/sfepy/discrete/materials.py
+++ b/sfepy/discrete/materials.py
@@ -93,6 +93,15 @@ class Material(Struct):
     def __init__(self, name, kind='time-dependent',
                  function=None, values=None, flags=None, **kwargs):
         """
+        A material is defined either by a function, or by a set of constant values,
+        potentially distinct per region. Therefore, either `function` must be specified,
+        or a combination of `values` and `**kwargs`.
+
+        For constant materials, `**kwargs` are simply combined with `values` into a dictionary
+        mapping material parameter names to parameter values.
+        The parameter values may either be specified as a constant value, or as another dictionary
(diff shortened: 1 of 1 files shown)


rc:
I agree with the new code structure, thanks for the improvement! I have just a few remarks:

- Make sure the lines are no more than 80 characters long.
- The names of material parameters that start with the dot are special, meaning that those parameters are not evaluated in quadrature points. One can use those e.g. for some flags modifying a  term behavior or other geometry independent data. This is to distinguish from the standard material parameters (e.g. elasticity constants), that can have constant values, but potentially can be different in every quadrature point (qp). Hence also the `'qp'` and `'special'` modes in the material evaluation functions. (The third mode `'special_constant'` might disappear - I have used it in my past projects but not anymore.) That is why `ConstantFunctionByRegion` ignores parameters with names starting with a dot - you inferred correctly that there the values should be dicts (`{region_name1 : value1, ..., region_nameN : valueN}`)

sfepy/discrete/materials.py
```diff
@@ -110,42 +119,49 @@ def __init__(self, name, kind='time-dependent',
...
-                  % self.name
+        if kwargs:
+            if values is None:
+                values = kwargs
+            else:
+                values = dict(values)
+                values.update(kwargs)
+
+        if (function is not None) and (values is not None):
+            msg = f'material can have function or values but not both! ({self.name})'
```
rc, line 130: Let us use here also the `f'material {self.name}: ...` form of the message. E.g.
```
msg = f'material {self.name}: specify function or values but not both!'
```

rc:
Ping @burnpanck - just checking that you got the above remarks.

Issue #743, opened 2021-11-10
unable to convert 2d mesh made with gmsh


flothesof:
Hi,

I am trying to model a 2D problem using GMSH. The mesh is bounded by a rectangle and in the shape of a car.
![image](https://user-images.githubusercontent.com/2360669/141091333-2e4a7b18-4f29-490a-be8a-caf596342424.png)
To use it within my 2D model, I first convert it following the instructions in the documentation.
```python ..\sfepy\script\convert_mesh.py --2d citroen.msh citroen.h5```
However, the conversion script fails with an error:
```python_traceback
sfepy: reading mesh (citroen.msh)...
Traceback (most recent call last):
  File "..\sfepy\script\convert_mesh.py", line 237, in <module>
    main()
  File "..\sfepy\script\convert_mesh.py", line 162, in main
    mesh = Mesh.from_file(filename_in)
  File "c:\users\fl232714\documents\python workspace\scripts\2020_sfepy\sfepy\sfepy
\discrete\fem\mesh.py", line 220, in from_file
    mesh = io.read(mesh, omit_facets=omit_facets)
  File "c:\users\fl232714\documents\python workspace\scripts\2020_sfepy\sfepy\sfepy
\discrete\fem\meshio.py", line 393, in read
    cgdata0[m.cell_sets[k][ic]] = cgrps[ik]
IndexError: index 30 is out of bounds for axis 0 with size 1
```
I’ve attached a debugger to the file and tried to understand why this happens.
```python
            for ic, c in enumerate(m.cells): #iterate over cells
                cgdata0 = nm.zeros((len(c.data),), dtype=nm.int32)
                for ik, k in enumerate(keys):
                    cgdata0[m.cell_sets[k][ic]] = cgrps[ik] # this fails since the index is not valid for cgdata0
                cgdata.append(cgdata0)
```
It turns out that `m.cell_sets[k][ic]` is a 2d numpy array in certain cases which leads to this bug:
```
m.cell_sets[k][ic]    
Out[2]: array([30, -5], dtype=int32)
```
Do you have any ideas why this is so and what I can do to prevent that? I’m guessing that I somehow have wrongly defined things in my script, but what?

Cheers,
Florian
Attached: .geo and .msh in archive
[car.zip](https://github.com/sfepy/sfepy/files/7511731/car.zip)

flothesof:
An even simpler example that throws the same type of error is that of a square in a square:
[...]

PR #744, opened 2021-11-11
new advice: use msh22 instead of msh4

vlukes:
This PR is inspired by #743.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -128,6 +128,12 @@ Mesh-Related Tasks
    ``<dimension>`` is the topological dimension of cells that should be in the
    mesh. For example, ``-d 2`` stores only the 2D cells.
 
+#. It is suggested to use ``msh22`` format instead of the default ``msh4``
+   when generating a mesh with ``gmsh``::
+
+      gmsh -2 cylinder.geo -o cylinder.msh -format msh22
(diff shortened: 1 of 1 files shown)


rc:
Thanks! What about adding also an information about why it is a good idea to use a particular gmsh format?

doc/faq.rst
```diff
@@ -128,6 +128,10 @@ Mesh-Related Tasks
    ``<dimension>`` is the topological dimension of cells that should be in the
    mesh. For example, ``-d 2`` stores only the 2D cells.
 
+#. Use ``msh22`` format instead of the default ``msh4`` when generating a mesh 
+   with ``gmsh``::
```
rc, line 132: (the ``msh4`` format does not work well with the -2 option)??

vlukes: OK, I would write "It is suggested to use ..., msh22 is more reliable and foolproof."

rc: OK!

vlukes: Updated.

rc: LGTM, just fix the line length, please :)

vlukes: Fixed :)

Issue #374, opened 2017-03-16
docs: link terms with the example(s)


rc:
It would be nice to have a list for each term of links to examples, where the term is used.

rc:
On 10. 03. 21 15:47, Antony wrote:
> Hey,
> If this issue is still active, It would be a pleasure to help.
> 

Yes, that would be welcome! FWIW: There are a script that lists terms use in 
examples (script/show_terms_use.py) and a script that generates the term table 
document (script/gen_term_table.py).

DuckNrOne:
Great! And the task consists of adding a new column to each [term table](http://sfepy.org/doc-devel/terms_overview.html#table-of-basic-terms) and each cell includes links to examples, that uses the term in the same row? Sounds like a job for me :D

rc:
Yes, something like that. I am not sure about the PDF version of the docs, let's see how it goes.  :)

DuckNrOne:
The PDF shouldn't be a problem, when I edit `gen_term_table` and the generation process of the PDF accesses this module in any way. Does it?

rc:
Technically no problem, it is automatic (sphinx does all the work), but the PDF pages are not so wide/flexible as the html docs pages. The term table in the [manual](http://sfepy.org/doc/_downloads/63e3f3e0a0635ddc67af3c6e4d0f6b5f/sfepy_manual.pdf) is already rather cluttered, so I thought that maybe the links to the examples can be only in the html version of the docs.

DuckNrOne:
I'll look into the module the next days if it possible to choose the columns of the generated table eg in the method call.

rc:
OK, anyway, my concern is rather premature - maybe it will look fine in the pdf too - ignore it for now :)

antonykamp:
I have a small question: Compiling the docs returns a warning (more an error) because sphinx cannot find [`examples/index`](https://github.com/sfepy/sfepy/blame/master/doc/examples.rst#L17). It could be possible that maybe autodoc doesn't work correctly in the toctree. Do you know why?
 More importantly: Because I would refer to the examples the same way, I don't know which links I should use given the name of an example. At the moment, I use [`examples/[example-name]`](https://github.com/antonykamp/sfepy/blob/e83e5ab12dcfdfa447e44cbd8b26a53061792fc5/script/gen_term_table.py#L133) :)

rc:
[...]

PR #751, opened 2021-11-18
2d fluid flow on car mesh

flothesof:
Hi,

as discussed in https://github.com/sfepy/sfepy/issues/743, here is a PR to add this example to sfepy.

I have a few questions:
- is it possible to tell resview how to add streamlines to the plot? I wrote a pyvista script to animate this example, which could also be part of the file (see results here https://twitter.com/frolianlb/status/1460231767542190083)
- should I add the .geo script to the repo? if so, how do I document its existence in the text?
- is the mesh too big for this kind of example? (2MB)

Cheers,
Florian


rc:
Hi Florian,

thanks, looks good! I have only some remarks below.

> I have a few questions:
> 
>     * is it possible to tell resview how to add streamlines to the plot? I wrote a pyvista script to animate this example, which could also be part of the file (see results here https://twitter.com/frolianlb/status/1460231767542190083)

Nice animation! IMO resview cannot add streamlines now (cc-ing @vlukes), but adding the visualization code to the example would work well.
 
>     * should I add the .geo script to the repo? if so, how do I document its existence in the text?

You can document  the mesh generation in the docstring.

>     * is the mesh too big for this kind of example? (2MB)

It is OK,  but if you could get reasonable results with a smaller mesh it might be better.
 
Now the remarks:

- Make sure, please, that also the lines in the docstring are no more than 80 characters long.
- Consider using `\nabla` for the gradient and `\nabla \cdot` for the divergence like in the other examples.
- The second dosctring equation has missing `}`, use:
```
    \int_{\Omega} \nabla \psi \cdot \nabla \phi
    = \int_{\Gamma_{left}} \ul{v}_0 \cdot n \psi
    + \int_{\Gamma_{right}} \ul{v}_0 \cdot n \psi
    \;, \quad \forall \psi \;,

where :math:`\ul{v}_0` is ....
```
and define v_0.
- There are no Dirichlet BCs, so the matrix is singular. The `'ls.scipy_direct'` (UMFPACK) solver is robust so it only complains, but other solvers might simply fail. Try fixing the potential in a single vertex, for example:
```
regions = {
    'Omega' : 'all',
    'Gamma_Left' : ('vertices in (x < 0.1)', 'facet'),
    'Gamma_Right' : ('vertices in (x > 1919.9)', 'facet'),
    'Vertex' : ('vertex in r.Gamma_Left', 'vertex'),
}
...
ebcs = {
    'fix': ('Vertex', {'phi.0': 0.0}),
}
```

flothesof:
I’ve slightly extended the example (possibility of taking into account any far field v0) and rewritten the equations as you suggested. 

@vlukes do you have any thoughts about the streamlines? they are available in pyvista so maybe this could be integrated into resview.py?

examples/diffusion/laplace_fluid_2d.py
```diff
@@ -0,0 +1,103 @@
...
+"""
+from sfepy import data_dir
+
+filename_mesh = data_dir + '/meshes/2d/citroen.h5'
+
+v0 = np.array([1, 0.25])
+
+materials = {
+    'm': ({'sigma': 1.0},),
+    'f': ({'left': np.dot(v0.T, np.array([-1, 0])),
```
rc, line 57: NumPy import is missing. (Also consider using `nm` like most other examples do (historical reasons)).

examples/diffusion/laplace_fluid_2d.py
```diff
@@ -0,0 +1,103 @@
...
+
+integrals = {
+    'i': 2,
+}
+
+equations = {
+    'Laplace equation':
+        """dw_laplace.i.Omega( m.sigma, psi, phi )
+         = dw_integrate.i.Gamma_Left(f.left, psi)
+         + dw_integrate.i.Gamma_Right(f.right, psi)"""
```
rc, line 94: The top and bottom regions and material are not used?

examples/diffusion/laplace_fluid_2d.py
```diff
@@ -0,0 +1,103 @@
...
+    python3 resview.py citroen.vtk
+
+
+Generating the mesh
+-------------------
+
+The mesh can be generated with::
+    gmsh -2 -f msh22 meshes/2d/citroen.geo -o meshes/2d/citroen.msh
+    python script/convert_mesh.py --2d meshes/2d/citroen.msh \
+        meshes/2d/citroen.h5
```
rc, line 47: Use  `python3`. It is also OK to have the command line examples longer than 80 chars - a single line is easier to copy & paste.

rc:
Thanks for the update! I have realized that you might want to use `dw_surface_ndot` term - it would multiply by the normal automatically.

vlukes:
> @vlukes do you have any thoughts about the streamlines? they are available in pyvista so maybe this could be integrated into resview.py?

Hi @flothesof! Yes, I will try to implement streamlines into `resview.py`.

examples/diffusion/laplace_fluid_2d.py
flothesof: I’ll add it (I forgot to re-test after copying from my main script)

examples/diffusion/laplace_fluid_2d.py
flothesof: Please see my next version (I forgot to re-test after copying from my main script) where the domains are used :)

examples/diffusion/laplace_fluid_2d.py
flothesof: I’ll make a longer line then!

rc:
Thanks @flothesof, almost there! Just two(three) things to settle:
- Use `python3` as the executable name in the docstring (line 46).
- Note that you can use `dw_surface_ndot` term - using it would make it easier for the normals not aligned with coordinate axes.
- We can either wait to have the streamlines in resview (thanks @vlukes!), or merge the example as it is and update it later.

vlukes:
> We can either wait to have the streamlines in resview (thanks @vlukes!), or merge the example as it is and update it later.

I suggest to merge the example as it is. I will implement the streamlines to `resview` later.

flothesof:
I used the `dw_surface_ndot ` term, as far as I could tell it there wasn’t any example using it so far.
Executing the script should look like this:
image

rc:
> I used the `dw_surface_ndot ` term, as far as I could tell it there wasn’t any example using it so far. @rc are the examples automatically linked to the terms table page? I find it very useful, as a beginner user of Sfepy, to be able to look at an example featuring the terms I’m trying to use.

Yes, the links are automatically generated (PR #374). They are meant exactly for this purpose :)

Thanks for the update, LGTM! I will merge it now, and later we can add the instructions how to display the streamlines.

Maybe attach your visualization script to this PR, so that we can take inspiration for the resview update?

rc:
Thanks for your contribution!

flothesof:
You’re welcome :)

flothesof:
Here’s a sample animation script that I have used (ping @vlukes for streamlines computed with the gradient):
```python
# load the mesh
import pyvista as pv
import numpy as np

mesh = pv.read("citroen.vtk")

# add gradients
grad = mesh.compute_derivative(scalars="phi")

# get min max coords of mesh
minimaxi = np.c_[mesh.points.min(axis=0), mesh.points.max(axis=0)][:2, :]


def make_points(minimaxi, n_points=100):
    points = np.zeros((n_points, 3))
    for coord in range(2):
        low, high = minimaxi[coord, 0], minimaxi[coord, 1]
        points[:, coord] = np.random.randint(low=low, high=high, size=n_points)
    return points

# and now an animation
points = make_points(minimaxi, n_points=10000)
point_cloud = pv.PolyData(points)
point_cloud = point_cloud.interpolate(grad, null_value=0, radius=20.)

streamlines = grad.streamlines(
    pointa=(0.1, 0, 0),
    pointb=(0.1, minimaxi[1, 1], 0),
    n_points=20,
    max_time=2000.,
    compute_vorticity=False,  
)


plotter = pv.Plotter()
plotter.add_mesh(grad, show_scalar_bar=False)
plotter.add_mesh(streamlines, show_scalar_bar=False)
plotter.add_mesh(point_cloud, scalars="phi", show_scalar_bar=False, colormap="inferno", opacity=0.5)
plotter.view_xy()
plotter.camera_position = [(868.4118785064393, 500.3989374701676, 2417.938072036079),
 (868.4118785064393, 500.3989374701676, 50.0),
 (0.0, 1.0, 0.0)]
plotter.open_movie("citroen.mp4")

for i in range(500):
    print(i, interpolated['gradient'].min(), interpolated['gradient'].max())
    interpolated = point_cloud.interpolate(grad, null_value=0, radius=20, strategy='mask_points')
    points = interpolated.points + 10.5 * interpolated["gradient"]
    points[:, 0] = points[:, 0] % minimaxi[0, 1]
    points[:, 1] = points[:, 1] % minimaxi[1, 1]
    plotter.update_coordinates(points, render=False)
    point_cloud = pv.PolyData(points)
    plotter.render()
    plotter.write_frame()
plotter.show()
plotter.close()
```

PR #752, opened 2021-11-23
New: allow to reinitialize the homogenization engine

vlukes:
I have a homogenization problem and I need to change the mesh in the parametric hook. To do this, I create a new problem and I need to reinitialize the homogenization engine in `HomogenizationApp`. This can be done by this PR and by setting `new_problem.force_init_he = True` in the parametric hook.

Or do you have another idea how to change the behaviour of the "parent" application in the parametric hook where only the problem instance is available?

### DIFF
--- a/sfepy/homogenization/homogen_app.py
+++ b/sfepy/homogenization/homogen_app.py
@@ -203,7 +203,9 @@ class HomogenizationApp(HomogenizationEngine):
 
         ret_all = get_default(ret_all, opts.return_all)
 
-        if not hasattr(self, 'he'):
+        force_init_he = hasattr(self.problem, 'force_init_he')\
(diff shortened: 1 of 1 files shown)


rc:
I have not used parametric hooks since a long time - maybe  you can parameterize directly the `HomogenizationApp` instead of the problem? Anyway, I have no other solution right now - I would need to see the problem definition to be able to say more. -> LGTM, thanks! Let me know whether you search another solution or wish to merge it as-is.

(FWIW: recently I tend to use `()` to enclose a long line that needs wrapping instead of the line continuation character. Seems nicer :))

vlukes:
Please, merge it as is. This solution is simple and straightforward.

PR #754, opened 2021-11-25
Resview streamlines

vlukes:
This PR adds streamlines option to `resview.py` as discussed in #751. Thanks to @flothesof for his sample script!

### DIFF
--- a/resview.py
+++ b/resview.py
@@ -584,8 +584,8 @@ helps = {
         'define position of plot labels [default: "225,75,0.9"]',
     'step':
         'select data in a given time step',
-    'xy_view':
-        'view the XY plane',
+    '2d_view':
+        '2d view of XY plane',
 }
 
 
@@ -654,9 +654,9 @@ def main():
     parser.add_argument('--off-screen',
                         action='store_true', dest='off_screen',
                         default=False, help=helps['off_screen'])
-    parser.add_argument('-x', '--xy-view',
-                        action='store_true', dest='xy_view',
-                        default=False, help=helps['xy_view'])
+    parser.add_argument('-2', '--2d-view',
+                        action='store_true', dest='view_2d',
+                        default=False, help=helps['2d_view'])
 
     parser.add_argument('filenames', nargs='+')
     options = parser.parse_args()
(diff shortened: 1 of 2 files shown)


resview.py
```diff
@@ -633,6 +654,9 @@ def main():
     parser.add_argument('--off-screen',
                         action='store_true', dest='off_screen',
                         default=False, help=helps['off_screen'])
+    parser.add_argument('-x', '--xy-view',
```
rc, line 657: Maybe remove the single-letter option? It seems ambiguous.

rc:
Great, thanks! I am wondering whether `t` is the best for the streamlines, but cannot come up with a better letter. :) (l?)
Just one question above, LGTM.

rc:
BTW. it would be nice to update the `gen_gallery,py` script to use resview and plot the streamlines for this example. A topic for a next PR.

resview.py
vlukes: Yes, It can be confusing, but `--xy-view` is rather long and `-x` does not collide with anything else. So I would keep it. It is not the best, like `t` is not, but I do not have a better suggestion.

rc: OK (but see below).

Note that you need not to write the complete option name if it does not coincide with another option - so writing `--xy` would work.

Moreover, argparse allows specifying a number of variants rather freely, so the first/short option version could be `--xy` or even `-xy` explicitly. Not so long anymore :)

vlukes: So what about using `--plane_xy_view` and shorten to `-p`?

rc: `--plane-view`, `-p`? It depends if you would like to add other view options in future (xz, isometric, parallel projection, whatever pyvista can). The convert_mesh.py script has `-2`, `--2d` in a similar setting. Well, sorry for the bike-shedding, yours will be the last word :)

vlukes: Finally, `-2` and `--2d` is probably the best.

vlukes: .. or better: `--2d-view`

rc: OK!

rc:
LGTM, merging...

PR #765, opened 2022-01-09
Fix for python3.10

rmcgibbo:
collections.Sequence -> collections.abc.Sequence
See for example: https://github.com/wireservice/agate/pull/737/files


rc:
Thanks, LGTM! (Although it is currently not tested here - see #766.)

Issue #773, opened 2022-01-25
Simplification of term table?


vlukes:
It seems to me that the term table is too complicated.

I think it is unnecessary to have separated integrals for `virtual, state` and `parametr_1, parametr_2` in the table.
If we write that `v`, `s`, ...  "_are test or parameter function_" description of many terms can be simplified without losing information.

**Example** - `dw_div_grad` term: there could be just two integrals in the definition,  with and without material parameter, and 

  `opt_material`,  `virtual/parameter_1`, `state/parameter_2`

in the column for arguments.

rc:
True, most new terms  (e.g. multilinear ones) allow different kinds of variables, so the distinction is no longer necessary.

PR #774, opened 2022-01-26
New sensitivity terms

vlukes:
This PR adds new multilinear terms:

 - sensitivity terms: `de_sd_diffusion`, `de_sd_div_grad`, `de_sd_dot`, `de_sd_lin_elastic`, `de_sd_piezo_coupling`,  `de_sd_stokes`
 - other terms: `de_diffusion`, `de_nonsym_elastic`

New sensitivity terms supports `eval` and `weak` modes. 

The term docstrings are written in a simplified way as suggested in #773.


rc:
Thanks!  I will have a look soon. For the moment some food for thought:

- We could either put the (new and old) multi-linear terms into the files of the original terms, or we could end up with all the terms in a single file. Not sure what is better, maybe using several files to group related terms together.
- The sensitivity terms do a number of calculations outside the final einsum function. `ETermBase` could be modified to allow sums of several einsum expressions passed into `.make_function()`, but I wonder what else could be added to facilitate writing the sensitivity terms and to be able to exploit the backends and their optimizations. The mesh velocity variable is a FE field, right?

vlukes:
> Thanks! I will have a look soon. For the moment some food for thought:
> 
>     * We could either put the (new and old) multi-linear terms into the files of the original terms, or we could end up with all the terms in a single file. Not sure what is better, maybe using several files to group related terms together.

Maybe, we can start by separating the multi-linear sensitivity terms into a new file, eq. `terms_multilinear_sensitivity`. What do you think? 

>     * The sensitivity terms do a number of calculations outside the final einsum function. `ETermBase` could be modified to allow sums of several einsum expressions passed into `.make_function()`, but I wonder what else could be added to facilitate writing the sensitivity terms and to be able to exploit the backends and their optimizations. The mesh velocity variable is a FE field, right?

Right, the mesh velocity is a FE field. I was thinking about using `einsum()` and `make_function()` for all calculations around, but the current implementation seems to be clearer to me, and therefore it is easier to debug it, and the performance is not so bad comparing to old `ev_sd_*` terms.

My motivation is in implementation of one specific shape sensitivity problem involving the shape sensitivity terms in the weak formulation, but `ev_sd_*` terms do not allow this.

rc:
> > Thanks! I will have a look soon. For the moment some food for thought:
> > ```
> > * We could either put the (new and old) multi-linear terms into the files of the original terms, or we could end up with all the terms in a single file. Not sure what is better, maybe using several files to group related terms together.
> > ```
> 
> Maybe, we can start by separating the multi-linear sensitivity terms into a new file, eq. `terms_multilinear_sensitivity`. What do you think?

Just `terms_sensitivity.py` would be fine.
 
> > ```
> > * The sensitivity terms do a number of calculations outside the final einsum function. `ETermBase` could be modified to allow sums of several einsum expressions passed into `.make_function()`, but I wonder what else could be added to facilitate writing the sensitivity terms and to be able to exploit the backends and their optimizations. The mesh velocity variable is a FE field, right?
> > ```
> 
> Right, the mesh velocity is a FE field. I was thinking about using `einsum()` and `make_function()` for all calculations around, but the current implementation seems to be clearer to me, and therefore it is easier to debug it, and the performance is not so bad comparing to old `ev_sd_*` terms.

OK.

I noticed that you are using `dot_sequences()` - what is its speed comparing to an optimized `einsum()` call? 

I wonder how difficult it would be to derive the shape derivative einsum expression symbolically just from the original term einsum expression. It could be a nice project :)

Is it possible to write (almost) all SD terms with the original term expression but a different "material" involving the velocity field?

> 
> My motivation is in implementation of one specific shape sensitivity problem involving the shape sensitivity terms in the weak formulation, but `ev_sd_*` terms do not allow this.

I see.

vlukes:
> Is it possible to write (almost) all SD terms with the original term expression but a different "material" involving the velocity field?

It is not possible to do this for all SD terms. E.g. the linear elasticity term must be expressed using the non-symmetric gradients in the SD version, `p div(v)` --> `p X(V) grad(v)`, etc.

vlukes:
> I noticed that you are using `dot_sequences()` - what is its speed comparing to an optimized `einsum()` call?

According to my observations, `dot_sequences()` is usually 10 times faster than `opt_einsum.contract()` or `nm.einsum(..., optimize='greedy')`.

sfepy/terms/terms_multilinear.py
```diff
@@ -1096,7 +1139,7 @@ class EIntegrateOperatorTerm(ETermBase):
     :Definition:
 
     .. math::
-        \int_\Omega q \mbox{ or } \int_\Omega c q
+        \int_\cal{D} q \mbox{ or } \int_\cal{D} c q
```
rc, line 1142: Use `{\cal{D}}`...

sfepy/terms/terms_multilinear.py
```diff
@@ -1529,3 +1522,160 @@ def get_function(self, mat, parameter, mode=None, term_mode=None,
...
+
+    :Arguments 1:
+        - material : :math:`\ull{D}`
+        - virtual  : :math:`\ul{v}`
+        - state    : :math:`\ul{u}`
+
+    :Arguments 2:
+        - material    : :math:`\ull{D}`
+        - parameter_1 : :math:`\ul{w}`
+        - parameter_2 : :math:`\ul{u}`
```
rc, line 1547: Merge the arguments?

sfepy/terms/terms_multilinear.py
```diff
@@ -1529,3 +1522,160 @@ def get_function(self, mat, parameter, mode=None, term_mode=None,
...
+    arg_types = (('material', 'virtual', 'state'),
+                 ('material', 'parameter_1', 'parameter_2'))
+    arg_shapes = {'material': 'D2, D2', 'virtual': ('D', 'state'),
+                  'state': 'D', 'parameter_1': 'D', 'parameter_2': 'D'}
+    modes = ('weak', 'eval')
+
+    def get_function(self, mat, virtual, state, mode=None, term_mode=None,
+                     diff_var=None, **kwargs):
+        fun = self.make_function(
+            'IK,n(i:j)->I,n(k:l)->K', mat, virtual, state, diff_var=diff_var,
```
rc, line 1559: Is not this just `'ijkl,i.j,k.l'`, but `mat` needs to be reshaped to the fourth order? The reshaping could be done without a copy, right? Manually here...

sfepy/terms/terms_multilinear.py
```diff
@@ -1529,3 +1522,160 @@ def get_function(self, mat, parameter, mode=None, term_mode=None,
...
+
+
+class ELinearTractionTerm(ETermBase):
+    r"""
+    Linear traction term. The material parameter can have one of the
+    following shapes:
+
+        - 1 or (1, 1) - a given scalar pressure
+        - (D, 1) - a traction vector
+        - (S, 1), (2D, 1) or (D, D) - a given stress in symmetric or
```
rc, line 1620: What is `(2D, 1)`?

sfepy/terms/terms_multilinear.py
```diff
@@ -1529,3 +1522,160 @@ def get_function(self, mat, parameter, mode=None, term_mode=None,
...
+            elif tdim == dim and tdim2 == dim:  # traction tensor - full
+                force = dot_sequences(traction, nv, 'AB')
+            elif tdim == dim**2 and tdim2 == 1:  # traction tensor - full, vector
+                aux = traction.reshape((-1, n_qp, dim, dim))
+                force = dot_sequences(aux, nv, 'AB')
+            elif tdim == sym and tdim2 == 1:  # traction tensor - symetric
+                aux = sym2nonsym(traction, [2]).reshape((-1, n_qp, dim, dim))
+                force = dot_sequences(aux, nv, 'AB')
+            else:
+                raise NotImplemented
```
rc, line 1675: `NotImplementedError('unsupported material shape! ({})'.format((tdim, tdim2)))` - is not this caught by `arg_shapes`?

sfepy/terms/terms_sensitivity.py
```diff
@@ -0,0 +1,392 @@
...
+    :Arguments 1:
+        - material    : :math:`g_{kij}`
+        - virtual/parameter_v : :math:`\ul{v}`
+        - state/parameter_s : :math:`p`
+        - parameter_mv : :math:`\ul{\Vcal}`
+
+    :Arguments 2:
+        - material : :math:`g_{kij}`
+        - state    : :math:`\ul{u}`
+        - virtual  : :math:`q`
```
rc, line 94: Can be also parameters?

sfepy/terms/terms_sensitivity.py
```diff
@@ -0,0 +1,392 @@
...
+          - {\partial \Vcal_j \over \partial x_i}
+
+    :Arguments 1:
+        - virtual/parameter_v: :math:`\ul{v}`
+        - state/parameter_s: :math:`p`
+        - parameter_mv: :math:`\ul{\Vcal}`
+
+    :Arguments 2:
+        - state    : :math:`\ul{u}`
+        - virtual  : :math:`q`
```
rc, line 193: Can be also parameters?

sfepy/terms/terms_surface.py
```diff
@@ -113,7 +113,7 @@ class SDLinearTractionTerm(Term):
     :Definition:
 
     .. math::
-        \int_{\Gamma} \ul{v} \cdot \ull{\sigma} \cdot \ul{n},
+        \int_{\Gamma} \ul{v} \cdot (\ull{\sigma}\, \ul{n}),
```
rc, line 116: `\cdot` is at many places, is this needed? It is not limited to dot of vectors in the docs...

rc:
> > I noticed that you are using `dot_sequences()` - what is its speed comparing to an optimized `einsum()` call?
> 
> According to my observations, `dot_sequences()` is usually 10 times faster than `opt_einsum.contract()` or `nm.einsum(..., optimize='greedy')`.

I expected it to be faster, but not by as much. We could try generating code that uses `matrix_multiply()`. :)

rc:
Would you mind using a smaller mesh for the `tests/test_term_sensitivity.py`? Currently, `mprof run run_tests.py tests/test_term_sensitivity.py` ([memory_profiler](https://github.com/pythonprofilers/memory_profiler)) reports peak memory usage 2.5GB, which kills my laptop. Also consider adding `gc.collect()` at the end of the inner loop of test_sensitivity() - this helps reducing the peak to 1.7 GB, as well as the overall usage :).

sfepy/terms/terms_multilinear.py
```diff
@@ -1418,28 +1426,20 @@ class EStokesTerm(ETermBase):
...
-        - virtual  : :math:`\ul{v}`
-        - state    : :math:`p`
+        - material: :math:`c` (optional)
+        - virtual/parameter_v: :math:`\ul{v}`
+        - state/parameter_s: :math:`p`
 
     :Arguments 2:
         - material : :math:`c` (optional)
         - state    : :math:`\ul{u}`
         - virtual  : :math:`q`
```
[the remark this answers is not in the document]
vlukes: It is obvious from `Arguments 1` that the first (non-material) argument is a vector field and the second one is a scalar field. Why to say the same in `Arguments 2`?

sfepy/terms/terms_multilinear.py
vlukes: Yes

sfepy/terms/terms_multilinear.py
vlukes: Yes

sfepy/terms/terms_multilinear.py
vlukes: It is my mistake.

sfepy/terms/terms_multilinear.py
vlukes: Yes, can be removed.

sfepy/terms/terms_surface.py
vlukes: I would use `\cdot` only for the scalar product. Otherwise, the equations will be hard to read.

sfepy/terms/terms_multilinear.py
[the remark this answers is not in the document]
rc: I though this was to indicate that parameter variables could be used in places of state/virtual. As it is now, it is not clear that arguments 2 support the `eval` mode. I prefer having arguments 1 and 2 separate - the meanings of the term and term-transposed differ in a system of PDEs.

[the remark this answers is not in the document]
vlukes: And what is the difference between `Arguments 1` = (`material`, `parameter_v`, `parameter_s`) and `Arguments 2` = (`material`, `parameter_v`, `parameter_s`) in the `eval` mode?

sfepy/terms/terms_surface.py
rc: IIRC it was used as "contraction" (matrix multiplication) as opposed to tensor product (`a \cdot b` having lower dimension as opposed to `a b` having higher). IMO `(,)` is also a notation for the dot product of "vectors" which is not the case here. The clearest might be just to use the Einstein summation convention and indices :) Anyway, revisiting the overall term notation is for another PR.
(Alternative: `\ul{v} \cdot \ull{\sigma} \ul{n}`)

vlukes:
> Would you mind using a smaller mesh for the `tests/test_term_sensitivity.py`?

Yes, we can use `cube_sphere` (6797 cells, 1337 nodes) instead of `cube_cylinder` (54517 cells, 10490 nodes).

sfepy/terms/terms_multilinear.py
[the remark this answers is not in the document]
rc: No difference, it just looks non-symmetric in the table :)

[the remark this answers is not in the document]
vlukes: But vector/scalar parameters can not be swapped, while virtual/state fields can be. So it is "non-symmetric".

sfepy/terms/terms_multilinear.py
```diff
@@ -1556,7 +1551,7 @@ class ENonSymElasticTerm(ETermBase):
     def get_function(self, mat, virtual, state, mode=None, term_mode=None,
                      diff_var=None, **kwargs):
         fun = self.make_function(
-            'IK,n(i:j)->I,n(k:l)->K', mat, virtual, state, diff_var=diff_var,
+            'IK,v(i:j)->I,v(k:l)->K', mat, virtual, state, diff_var=diff_var,
```
rc, line 1554: `v(i.j)`, because `:` is reserved for the symmetric gradient. So some tweaks will be needed in the parsing code.

vlukes: > So some tweaks will be needed in the parsing code.

Yes, I already know that :)

vlukes: > `v(i.j)`, because `:` is reserved for the symmetric gradient. So some tweaks will be needed in the parsing code.

Can be `:` used outside `s()`?

vlukes: Can be `s()` used without `:`?

rc:
Thanks, LGTM!

I noticed that in the generated term table the argument lists are compacted, but e.g. in dw_laplace, dw_diffusion still both integrals are there. I assume this is correct, as you updated only the sensitivity terms in this PR, right? If you want, you can rebase, but it is not necessary, the merge is clean as it is.

sfepy/terms/terms_multilinear.py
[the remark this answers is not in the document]
vlukes: > I noticed that in the generated term table the argument lists are compacted, but e.g. in dw_laplace, dw_diffusion still both integrals are there. I assume this is correct, as you updated only the sensitivity terms in this PR, right? 

Yes, only the argument lists of multilinear and sensitivity terms are changed in this PR.

> If you want, you can rebase, but it is not necessary, the merge is clean as it is.

Please, merge it as it is.

rc (approved):

PR #780, opened 2022-02-18
fix Problem.solve() initial state argument handling

rc:


examples/linear_elasticity/dispersion_analysis.py
```diff
@@ -307,7 +307,7 @@ def compute_von_mises(out, pb, state, extend=False, wmag=None, wdir=None):
 def save_eigenvectors(filename, svecs, wmag, wdir, pb):
     if svecs is None: return
 
-    variables = pb.get_variables()
+    variables = pb.set_default_state()
```
vlukes, line 310: It is not intuitive to use `set_default_state()` to get the problem variables.

rc: I could not come up with a better name, this came as an analogy to `dict.setdefault()`, which sets a default item (here creates and initializes the variables.vec attribute) and then returns the item (here the whole variables, not just variables.vec). Do you have an idea? Originally I had `.init_state()`, but that was not playing well with `Variables.init_state() which does not return anything. `.get_variables()` is already used for something else, and having `get` in a function for setting something is even weirder.

vlukes: I have no idea for a better name. :-/

rc: I have added a docstring that explains what it does.

PR #783, opened 2022-02-23
multilinear terms: use actual ndarray args

### LINKED ISSUE #782: Multilinear terms - cached values
The [attached](https://github.com/sfepy/sfepy/files/8117696/test_multilinear.zip) script shows the problem in multilinear terms.

There is the following equation:

`dw_diffusion.i.Omega(solid.K, q, p)  = de_sd_diffusion.i.Omega(solid.K, q, p, V)`

which has the sensitivity term `de_sd_diffusion` on the right hand side. The design velocity field `V` is changing in the defined parametric hook, but there is no change in the calculated value `p`!

rc:
This PR should fix #782.


rc:
@vlukes does it help also with your other examples?

vlukes:
> @vlukes does it help also with your other examples?

Yes, it seems to be working!

PR #790, opened 2022-03-08
use actual reference mapping data in get_einsum_ops()

rc:
- do not store the mapping data in ExpressionArg


vlukes (approved):
It seems working.

PR #793, opened 2022-03-09
remove homogen.py

rc:
Call `HomogenizationApp` from `simple.py`.


vlukes:
Yes, but running all simulations by calling `simple.py` is a little confusing. We have nothing like `complex.py`. So I suggest to rename `simple` to e.g. `solve` or `run_simulation`.  We have also `simple_homog_mpi.py` that should be renamed or modified as well as `simple.py`.

rc:
I would post-pone the renaming to when we move towards using setup.py entry points. Many scripts will be (re)moved and/or renamed, and the docs changes would be significant. As you can see in the diff, `homogen.py` was used almost nowhere, so removing it seems like a little change. (BTW. `simple.py` is named according to the first word in SfePy, it might have been `complex.py`, if the package was CfePy :))

vlukes:
Then why not post-pone removing `homogen.py` too? There are references to `homogen.py` and `simple.py` in several _Example Applications_, which are outside the main documentation in the separate repositories. So, it would be much easier to change it at once.

rc:
Forgot about those, added a proxy `homogen.py`.

vlukes (approved):

vlukes (approved):

Issue #552, opened 2019-09-06
Nixpkgs Sfepy recipe - fast test suite


wd15:
I'm trying to finish off a [PR for Sfepy in Nixpkgs](https://github.com/NixOS/nixpkgs/pull/67788). We've found that the full test suite is quite slow O(minutes). Could @rc recommend a subset of tests that is both representative and runs in O(seconds).

Would it be possible to add `--fast` flag to run_tests.py for a more permanent solution?

rc:
Yes. adding the `--fast` flag is a good idea. I guess most of the `test_input_*.py` tests can be omitted from the fast mode, need to think about other. Some modifications might be needed. I will try the flag next week.

wd15:
[Here](https://github.com/NixOS/nixpkgs/blob/6271f789126a74482f727ed58d54aae2bbdae9c7/pkgs/development/python-modules/sfepy/default.nix#L37) are the list of tests removed from the Nixpkgs Sfepy recipe to make things speedier.

rc:
Thanks for the list. I am wondering if, instead of improving the custom test runner, going with #363 would be better in the long run. Converting all the tests for use with pytest would be a lot of tedious work, but once it's done, things like marking slow tests would be easy.

wd15:
> Thanks for the list. I am wondering if, instead of improving the custom test runner, going with #363 would be better in the long run. Converting all the tests for use with pytest would be a lot of tedious work, but once it's done, things like marking slow tests would be easy.

Absolutely makes sense if the run_tests.py stuff is all going away soon. Regarding switching over to pytest, I would recommend providing a way for users to test an installed version of sfepy. I don't like the way virtually all Python packages require a clone in order to test. I've done this with [pymks](https://github.com/materialsinnovation/pymks/blob/develop/pymks/__init__.py#L19). Also put the tests directory as part of the distribution, but with no `__init__.py`.

rc:
As a matter of fact, `sfepy-run run_tests` allows running tests of the installed sfepy, so I would like to preserve this with pytest.

Issue #363, opened 2017-02-09
use PyTest for tests?


rc:
It looks like NumPy and SciPy are going to use http://doc.pytest.org/ for tests. Is it worth changing our testing framework?

heczis:
I am for changing the testing framework, since it would reduce the codebase and make use of a standard tool for testing (making it easier for people to transfer knowledge from other projects).
Personally, I would prefer `unittest`, since it is part of the standard library - this way we would not add a dependency.

rc:
For clarification: I created this issue based on the scipy mailing list discussion and [GSoC topic proposals](https://github.com/scipy/scipy/wiki/GSoC-2017-project-ideas). SciPy is already our dependency, so if they switch... This is not an urgent topic, but, of course, PRs (also proof-of-concept kind) are welcome! :)

deeptavker:
Hey! I'm new to open source and have been working on test driven development for a while. I recently did some work with pytest and I know the basic unittest.TestCase subclassing, assert statements and fixtures and CI. I'd like to take this up in case it is decided to shift to pytest.

rc:
@deeptavker thanks for the offer, we will ping you in case we decide going this way.

PR #803, opened 2022-04-11
Use pytest

rc:
This PR aims to address #363 and #552.

As a proof of concept, run:
```
pytest -sv tests/test_assembling.py
pytest -sv tests/test_base.py
```
I will convert more test files as time permits. @vlukes, @wd15, if you have some recommendations, please chime in. I will probably move the `tests/` directory into `sfepy/`, as mentioned in #552. The `setup.py` moves the tests there during install anyway.


wd15:
Just a note that you can ensure that the canonical way of running the tests is just "py.test" without addition arguments by adding the configuration to the setup.cfg as well as including and excluding tests. For example, pymks uses [this](https://github.com/materialsinnovation/pymks/blob/master/setup.cfg#L23).  Sfepy doesn't have a setup.cfg so might be handy to have one. Also, I'm not sure that setup.cfg is the latest and greatest way for python packaging, but it's what I've been using. It seems that the pyproject.toml is being used a lot.

rc:
Thanks @wd15!

rc:
The current source directory structure (only relevant shown) is
```
examples/
meshes/
sfepy/
tests/
```
while after the installation it is
```
sfepy/
...examples/
...meshes/
...tests/
```
With this setup, I am getting test failures (such as `ModuleNotFoundError: No module named 'examples.linear_elasticity'`) in examples that import other examples when testing the installed package.
I can "fix" that by adding `~/.local/lib/python3.8/site-packages/sfepy/` to `PYTHONPATH`, but that is not a solution.

So to fix it properly, I am considering moving both `tests/` and `examples/` under `sfepy/`, while leaving `meshes/` alongside `sfepy/` in the sources (those are OK thanks to `sfepy.data_dir` variable). I plan to use absolute imports, e.g. `from sfepy.examples.linear_elasticity.its2D_1 import ...` to avoid possible ambiguities in example names.

What do you think?

rc:
I guess it is almost ready. Any idea what causes the AppVeyor failure? Or would it be better to use github actions?

vlukes:
Would it be possible to use a more user-friendly way of calling the tests? E.g. `./simply.py --tests`? Or `./sfepy-run tests`?

rc:
Actually you can do (in the top source directory)
```
export PYTHONPATH=.
```
and then just
```
pytest sfepy/tests/
```
The way in the docs is a general one that works also for an installed sfepy version.

I would like to add a sfepy-test entry point in another PR, that would make sfepy pip-installable. Not sure when I get to it, though.

.appveyor.yml
```diff
@@ -38,9 +38,9 @@ environment:
   matrix:
 
     - PYTHON: "C:\\Python37_64"
-      PYTHON_VERSION: "3.7"
+      PYTHON_VERSION: "3.9"
```
vlukes, line 41: What is the reason to require Python 3.9?

vlukes (approved):

.appveyor.yml
rc: With 3.7 the tests did not work, so I tried updating. No help, it seems there are some problems related to using conda.

vlukes: This is a very strict requirement. The latest LTS Ubuntu 20.4 has Python 3.8.

rc: We can change it to 3.8, as it does not work anyway :)

vlukes (approved):

rc:
@wd15 I have marked several of the tests suggested as slow in #552 with the `slow` marker. I will merge this PR now as it is to move on, we can discuss further improvements in future PRs. Maybe you could use pytest's command line arguments to restrict the tests run to your exact needs?

wd15:
Great work @rc! What is the correct way to run the tests now? Would running

    $ python -c "import sfepy; sfepy.test()"

for an installed version run a fast version of the tests? What about for a cloned version?

rc:
Have a look [here](https://sfepy.org/doc-devel/installation.html#running-automated-test-suite). Let me know if the docs are not clear enough. I shamelessly copied the pymks `test()` function :)

vlukes:
I suggest adding a note on how to run a single test.

rc:
Thanks for the feedback, I tried to improve the docs in #814.

PR #805, opened 2022-04-12
Update ev div grad terms

vlukes:


rc:
Thanks! `python3 run_tests.py tests/test_term_call_modes.py` seems to fail on `de_grad`.

vlukes:
> Thanks! `python3 run_tests.py tests/test_term_call_modes.py` seems to fail on `de_grad`.

Yes, I see. But I'm not able to identify the reason.

rc:
I will have a look.

rc:
I see, the `de_grad` term cannot have the weak mode, unlike `dw_div`. The results of the weak integral must be a scalar and this is not the case here. That is also why there is no `dw_grad` term.

vlukes:
Thanks :)

rc (approved):

PR #834, opened 2022-06-16
Fix homog phononic

vlukes:


sfepy/homogenization/coefs_phononic.py
```diff
@@ -39,7 +39,9 @@ def compute_eigenmomenta(em_equation, var_name, problem, eig_vectors,
             eigenmomenta[ii, :] = 0.0
 
         else:
+            var.locked = False
             var.set_data(vec_phi.copy())
+            var.locked = True
```
rc, line 44: It might be better to use `variables.set_state_parts({var_name : vec_phi})`.

vlukes: Your fix has only one line of code, so it is definitely better :)

rc: But it does not work :] - try `python3 ./simple.py sfepy/examples/phononic/band_gaps.py --phonon-dispersion
`. One needs to call `variables.init_state()` at some time before calling `variables.set_state_parts()`.

The very original code works for me - you use it in some other example?

rc:
Thanks, LGTM!

PR #836, opened 2022-06-27
Fix gen_gallery.py + resview.py

rc:


rc:
The gallery looks OK, let's merge it. Polishing can come later.

PR #842, opened 2022-07-18
remove simple_homog_mpi.py

rc:
@vlukes, I have tested the PR with
```
mpiexec -n 4 python3 simple.py --app=bvp-mM --debug-mpi sfepy/examples/homogenization/nonlinear_hyperelastic_mM.py
```
It seems to work, besides some error messages after the computation ended, and `--debug-mpi` produces only an empty file. That was however also the case with `simple_homog_mpi.py`.


sfepy/examples/homogenization/nonlinear_hyperelastic_mM.py
```diff
@@ -1,7 +1,7 @@
 import numpy as nm
```
vlukes, line 1: It would be helpful to add 

`mpiexec -n 4 python3 simple.py --app=bvp-mM --debug-mpi sfepy/examples/homogenization/nonlinear_hyperelastic_mM.py`

hint to the docstring of this example?

vlukes (approved):

vlukes (approved):

Issue #320, opened 2016-04-19
add releases to PyPi


rc:
It would be nice to have a streamlined way of uploading releases to PyPi.

nschloe:
This.

From my experience, putting a project on pypi (and hence make it installable via a simlple `pip install sfepy`) gives any package a significant popularity boost.

PR #845, opened 2022-08-01
reorganize scripts

rc:
This PR aims to reduce the number of scripts and use entry points in setup.py with non-generic names (such as `simple.py`. The ultimate aim is `pip install sfepy` (#320).

It is still in an early stage, but you can try it using `pip install -e .` Comments/suggestions welcome.


sfepy/version.py
```diff
@@ -36,7 +36,7 @@ def get_basic_info(version=__version__):
     for up_dir in ['..', '.']:
         top_dir = op.normpath(op.realpath(op.join(op.dirname(__file__),
                                                   up_dir)))
-        aux = op.join(top_dir, 'README.rst')
+        aux = op.join(top_dir, 'VERSION')
```
vlukes, line 39: Where does the `VERSION` come from? When I clone `rc/reorganize-scripts-entry-points` and run `python3 setup.py build_ext --inplace` `VERSION` does not exist and the top level directory can not be determined.

rc: Good catch, fixed.

vlukes:
When running `make htmldocs` the `gen_term_table` module can not be found:

```
python3 setup.py htmldocs
running htmldocs
make[1]: Entering directory '/home/vlukes/tmp/sfepy/doc'
PYTHONPATH=.. sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.8.5

Extension error:
Could not import extension gen_term_table (exception: No module named 'gen_term_table')
make[1]: *** [Makefile:35: html] Error 2
make[1]: Leaving directory '/home/vlukes/tmp/sfepy/doc'
INFO: 
Using Python 3.8.
```

rc:
Should work now. The docs are not updated yet. I wonder how to replace `python simple.py ...` etc. in the docs - we could either use `python sfepy/scripts/simple.py ...` or assume an installed sfepy and use `sfepy-run ...`. Those two possibilities could be documented on a single place somewhere in the docs.

vlukes:
> Should work now. The docs are not updated yet. I wonder how to replace `python simple.py ...` etc. in the docs - we could either use `python sfepy/scripts/simple.py ...` or assume an installed sfepy and use `sfepy-run ...`. Those two possibilities could be documented on a single place somewhere in the docs.

This is a crucial question. Using `python sfepy/scripts/simple.py ...` to run simulations is really not user friendly. Is there a way to keep the top level startup script?

PR #861, opened 2022-09-16
update interface for mumps library version 5.4.x

vlukes:


rc:
It works, great! So the `mumps_struc_c` for 5.3 and 5.4 is the same?

vlukes:
Yes, `mumps_struc_c` is the same for 5.3 and 5.4.

PR #862, opened 2022-09-16
speed-up direct linear solver reuse in time-stepping solvers, unify presolve

rc:
Hey @vlukes, does this work for you? Now all options of `ls.auto_direct` handle the `presolve` parameter in the same way, and omitting the matrix digest check can significantly speed up time dependent problems. You can try it using the elastodynamic.py example (use larger mesh to see anything).


vlukes:
It seems to work for me.

PR #864, opened 2022-09-19
decrease source tarball size

rc:
Hey @vlukes, you might have some other ideas? With the current changes it goes from ~30MB to ~15MB.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ prune doc/_build
 prune sfepy/tests/.pytest_cache
 exclude doc/logo/*.vtk
 exclude doc/publish.py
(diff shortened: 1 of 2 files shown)


vlukes:
I think that 15MB is acceptable tarball size for the source.

Size of mesh files could be reduced by compressing the large files, e.g. `cube_big_tetra.mesh` has about 7MB while `cube_big_tetra.mesh.gz` has only 2MB. What do you think?

rc:
The 15MB are the gzipped sources, so zipping the files before making the tarball would not help. Anyway, I think it is OK too.

In another PR, we could revisit the meshes, remove the unused ones and possibly replace the "boring" ones, so that most of the examples are not on the cylinder or a block :)

PR #873, opened 2022-09-26
update evalauate_at(): return three-dimensional array

vlukes:


rc:
Thanks! I just wonder if adding an argument such as `keepdims` of `np.sum` might be preferable. But the way of this PR is definitely simpler. Let's merge it :)

Issue #876, opened 2022-09-29
Inconsistent gradient shape


vlukes:
The gradient shape returned by `terms_basic.dq_grad()` is (..., dim, n_comp) while `crefcoors.evaluate_in_rc()` gives (..., n_comp, dim).

According to [1], gradient of a vector field equals the transpose of the Jacobian matrix, which is given by

J_{ij} = \partial f_i  / \partial x_j

and has the shape of (n_comp, dim). So the gradient should have the shape (dim, n_comp).

But e.g. [2] says that the gradient of a vector field is organized into a single matrix known as the Jacobian matrix. See also this [3] discussion :)

[1] https://en.wikipedia.org/wiki/Gradient
[2] https://towardsdatascience.com/step-by-step-the-math-behind-neural-networks-d002440227fb
[3] https://math.stackexchange.com/questions/156880/what-does-it-mean-to-take-the-gradient-of-a-vector-field

rc:
Thanks for bringing this up. I would prefer, also based on the links you found, the `evaluate_at()` version, i.e. (n_comp, dim).

vlukes:
> I would prefer, also based on the links you found, the `evaluate_at()` version, i.e. (n_comp, dim).

I do not prefer this variant, because it means that the gradient of a scalar function is (1, dim), which is inconsistent with how the gradient is understood elsewhere in the code.

`dq_grad()` is also more wired in the code than `evaluate_at()`.

rc:
Yes, it might be more difficult. Note, however, that the current `FieldVariable.evaluate()` is internally inconsistent - the 'val' mode returns `(n_el, n_qp, n_comp, 1)`, while 'grad' returns `(n_el, n_qp, dim, n_comp)`. The 'div' and 'cauchy_strain' modes are IMO consistent with the 'val' mode.

 IIRC the origin of this "error" is connected with the FE bases, that have shape `bf.shape = (n_qp, 1, n_ep)` and `bfg.shape = (n_qp, dim, n_ep)`, to facilitate multiplication with element the DOF vector.

[...]

PR #877, opened 2022-09-30
Test gradients

vlukes:
Related to #876.  Are we able to pass this test? It would be nice.


vlukes:
Of course, the shape of `expected_values` must be adjusted according to conclusion of #876 (if any).

PR #881, opened 2022-10-12
sfepy-test

rc:
The new script aims at simplifying running tests - instead of
```
python -c "import sfepy; sfepy.test()"
```
only
```
sfepy-test
````
would be needed. Also passing pytest arguments will be trivial.

I still have to update the installation docs accordingly.


vlukes (approved):

vlukes:
Good idea!

PR #883, opened 2022-10-17
Fix phonon-plot

heczis:
In my case, this prevents two different axes with different scales being plot over each other.

### DIFF
--- a/sfepy/homogenization/band_gaps_app.py
+++ b/sfepy/homogenization/band_gaps_app.py
@@ -502,7 +502,7 @@ class AcousticBandGapsApp(HomogenizationApp):
                             bg.freq_range_initial,
                             plot_range,
                             show_legend=plot_opts['legend'],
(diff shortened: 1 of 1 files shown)


rc:
Thanks! It looks the same problems are also in `plot_dispersion()` - can you fix those too? But then the axes limits are not applied. 

As a matter of fact, the plotting functions need a facelift (the `new_axis` argument could be removed, taking axis as argument instead of `fig_num`, ...). 

So for now, your fix + apply the limits in `plot_logs()` would do, what do you think? It seems that adding `new_axes=True` to `plot_gaps()` calls instead of the `plot_logs()` calls would do it.

heczis:
I will look at it, perhaps together with #884.
But I have some deadlines approaching (say next 2 weeks), so give me some time, please :-]

rc:
Sure, take as much time as needed, no deadline here :)

PR #891, opened 2022-11-07
Fix simple phonon args

heczis:
Hi, here is my understanding of how simple.py should handle phononics:

* It should either do `band_gaps` or one or both of `phase_velocity` or `dispersion`.
* In addition, `plot` can make it to plot the results of `band_gaps` or `dispersion`.

The first issue is fixed by eventually raising a `RuntimeError` in case of conflicting parameters given by the user.

### DIFF
--- a/sfepy/examples/phononic/band_gaps_conf.py
+++ b/sfepy/examples/phononic/band_gaps_conf.py
@@ -231,6 +231,9 @@ class BandGapsConf(Struct):
         ema_options = copy(self.eigenmomenta_options)
         ema_options.update({'var_name' : 'u_c'})
 
+        dispersion_options = copy(self.band_gaps_options)
+        dispersion_options.update({'log_save_name' : 'dispersion.log'})
+
         coefs = {
             # Basic.
             'VF' : {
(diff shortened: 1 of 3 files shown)


rc:
Thanks!

What happens when `--phonon-dispersion --phonon-phase-velocity` are used together? It seems that then the phase velocity is not computed?

heczis:
Thanks for the hint. Forced me to actually read and understand the code, hopefully.
The last commit should make all the options work independently.
I decided to push first, test meanwhile, and perhaps merge the commits later, since the last one reverts some earlier changes.
Also, I thought about merging this branch with #883 (after I finish it). Or (more precisely): We could close #883 and include the [upcoming] changes in phonon plots here.

rc:
I found two minor issues:

 1. The x axis range in the plots is somehow too tight (the leftmost gap is at the figure edge for M and angle).
 2. When computing and not plotting - `python3 sfepy/scripts/simple.py sfepy/examples/phononic/band_gaps.py --phonon-dispersion  --phonon-band-gaps` - there is only one log, as both the 'band_gaps' and 'dispersion' homogenized "coefficients" share the `band_gaps_options` and so save into `band_gaps.log`.

heczis:
Thanks for the hints, as always ;-)
I added the first fix to the last of the previous commits. The other was fixed inside the band gaps example.

rc (approved):

rc:
LGTM!

PR #892, opened 2022-11-07
fix eig() method argument shadowing method parameter

rc:


rc:
I will merge it now as other things depend on it. Feel free to chime in in case of problems.

Issue #900, opened 2022-11-23
modeling heat loss with a Newton type term for diffusion terms?


flothesof:
Hello, 
related to my previous questions #898, I would like to use a term associated to Newton type heat loss on one of my boundaries (see Wikipedia [here](https://en.wikipedia.org/wiki/Newton's_law_of_cooling)).
This would involve adding a surface integral term of the type $\int_\Gamma h (T - T_\infty) d\Gamma$ to the set of equations.
I have searched the code base and the Github for related things but I couldn’t find any.
My question is: can I implement this with the existing terms? Or do I need to define an additional term for this?
I would be willing to contribute an example based on this if I could get it to work...
Thanks again for your help!
Florian

flothesof:
Sorry for the above post, it seems I was too quick to write... I just noticed that the term is actually listed in the docs https://sfepy.org/doc-devel/src/sfepy/terms/terms_dot.html#sfepy.terms.terms_dot.BCNewtonTerm
I’ll try to get it to work then.

rc:
Does it work then?

flothesof:
Yes, it seems to work as expected. Here’s a screenshot with the heat source on top and the heat loss at the bottom. It is a transient Poisson problem with three materials.
![image](https://user-images.githubusercontent.com/2360669/203940585-bc4e4704-e2be-41e0-9a6e-2b59b31e727b.png)

This could be turned into an example if you want.

rc:
Yes, nice examples are welcome! It could replace some of the more "boring" ones.

Issue #898, opened 2022-11-22
Transient thermal simulation on 3 material mesh, problems with Neumann source


flothesof:
Hi,
I’m currently trying to simulate the heat propagation on a tri-material mesh representing a metal plate on which a cylinder is built and surrounded by a low conductivity material.
I have used gmsh to draw, tag physical volumes and mesh the geometry. 
I export it to the .mesh (not .msh) format since the latter does not handle the material id parameters well (should I do a bug report on that?). Finally, I convert it to .vtk format using `sfepy-convert -d 3 .\extruded_mesh_with_cylinder.mesh .\extruded_mesh_with_cylinder.vtk`.
The result is this (featuring 4 materials, of which 2 belong to the same properties).
![image](https://user-images.githubusercontent.com/2360669/203318437-9ae82542-77ac-4148-a5ad-7f01f1fce1e9.png)
I have written a three volume model file (attached). It defines a surface boundary for a simple Neumann heat flux.
However, when I run the model, I find the result seems to be wrong since the heat source is not uniform on the top as I would have expected.
![image](https://user-images.githubusercontent.com/2360669/203319881-e2081d64-6bd4-4cdd-8890-f9330380e962.png)
Could you give me your opinion on what is going wrong? Or any tips for debugging? 
My best intuition right now is that it could have something to do with the way Gamma_Cylinder is defined (see code). I checked the regions for this but they look good to me.
This is basically a 3D version of #733. 
Many thanks!
Florian
[sfepy github issue.zip](https://github.com/sfepy/sfepy/files/10067555/sfepy.github.issue.zip)

flothesof:
I think I have found the issue : I confused some of the physical units used in my input file.
Correcting them as in the below gives good results:

```
materials = {
    'powder': ({'lam': 0.16 * 1e3,
                'rho_cp': 1330. * 650. * 1e-9 * 1e6},),
    'cylinder': ({'lam': 153. * 1e3,
                  'rho_cp': 2660. * 927. * 1e-9 * 1e6},),
    'plate': ({'lam': 153. * 1e3,
               'rho_cp': 2660. * 927. * 1e-9 * 1e6},),
}
```

PR #904, opened 2022-12-02
Transient poisson multimaterial example

flothesof:
Hi, as discussed in #900 and #898, I would like to submit this example.

I’ve developed it under Windows and had a number of issues during development of this example.
Most importantly, the documentation generation is broken under windows (and installation from source also).
So I couldn’t really see how the example is rendered.

What do you think?

Cheers,
Florian


rc:
Thanks, very nice!

Just a few notes:
- Format also the docstring to 80 chars line width.
- LaTeX chokes on the second equation - replace align -> aligned
- Add signs to terms in the equation :)
- The integral order two is OK as long as the material parameters are constant per element - it seems to be so here, right?
- Maybe replace `ls.scipy_direct` by `ls.auto_direct`, which can use MUMPS, if available. Also set a few options for linear problems:
```
    'ls' : ('ls.auto_direct', {
        # Reuse the factorized linear system from the first time step.
        'use_presolve' : True,
        # Speed up the above by omitting the matrix digest check used normally
        # for verification that the current matrix corresponds to the
        # factorized matrix stored in the solver instance. Use with care!
        'use_mtx_digest' : False,
    }),
```
is more than 4x faster on my laptop...

- Set `'i_max': 1,` in `nls.newton`.

In your application, do you plan to store a plot a time-history of the solution at some point in the domain? If yes, it would be nice also here.

flothesof:
Thank you for your comments.
I’ll try to come up with an improved version of the script.

Regarding the time-history, I am indeed using a probe for extraction at every time step using step_hook
I can try to integrate this because indeed this would be even nicer.

rc:
Thanks! I can do the formatting for you (well, my editor can ;)) - can I push to your branch? IMO github allows that.

flothesof:
Hi, yes, you can push to my branch! The little checkbox allowing this is checked :)

rc:
I have pushed my changes, so do not forget to fetch them :)

flothesof:
I have removed the call to plt.show (I found it a bit too convoluted to add the `define()` function just for this) and added the example to the list of testable examples.

rc:
Thanks, LGTM! I have just added a custom view for the gallery.

sfepy/tests/test_declarative_examples.py
```diff
@@ -25,6 +25,7 @@ def inedir(filename):
     'diffusion/sinbc.py',
     'diffusion/time_advection_diffusion.py',
     'diffusion/time_poisson.py',
+    'diffusion/time_heat_equation_multi_material.py',
```
rc, line 28: Put above time_poisson.py to preserve alphabetical order?

flothesof: Alright!

rc:
OK, let's merge it!

PR #908, opened 2022-12-08
fix resview.py for pyvista 0.37.0

rc:
Hi @vlukes, what do you think about this PR?

 There is another, not related, inconvenience: with `-e` also the glyphs have edges. Can this be solved easily? Anyway it's just a minor thing.


vlukes (approved):

rc:
Thanks. So I also removed the wireframe that was displayed with glyphs implicitly - now `-e` actually has some effect. Or?

vlukes:
Yes, perhaps it is better to show wireframe only when it is explicitly requested.

rc:
OK!

PR #919, opened 2022-12-31
resview.py: unify camera position handling in animation and normal modes

rc:
Hi @vlukes, this PR is sort of related to #918 - I struggled to get a correct camera setting for a 2D mesh, and on the way I made the changes presented here.

The still remaining problems are:
- The 'c' key does not work for meshes with no z component. (Hence #918 workaround.)
- The animation saving works only with --off-screen and for some inputs it leads to infinite
 ```
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
...
```
No idea why, maybe it's just my installation.

PR #920, opened 2023-01-09
fix the off-screen animation issue mentioned in #919

vlukes:


rc:
Thanks! With `--off-screen` the animation saving works now even for the input it did not. However, without `--off-screen`, the infinite exceptions still occur. I tried to comment out both show() and close() calls, but the results were the same. Can I send you the input so that you can try the same results?

vlukes:
Yes, send me the inputs and I will try it.

rc:
So it's my weird installation. Let's merge this, it makes at least off-screen working for me :)

PR #924, opened 2023-01-17
new MatchDOFsOperator: tie DOFs of two fields

vlukes:


rc:
Thanks, LGTM! Is there an example where we could demonstrate this condition?

vlukes:
There is an [example application](http://sfepy.org/sfepy_examples/example_shell_beam/). The condition is use to tie the displacements of shell10x and solid elements.

rc:
OK, I thought about something in sfepy, so that the code is tested by the test-suite, but I guess testing the `ShiftedPeriodicOperator` is sufficient. Or, the beam-shell example app is not that big - maybe you could include it in `sfepy/example/`?

Anyway, I will merge this.

rc (approved):

PR #928, opened 2023-01-20
resview.py: support saving animation as PNG sequence

rc:
Maybe also add --steps option to select which time steps to save?


vlukes (approved):
:+1:

Issue #879, opened 2022-10-04
Trouble installing SfePy on Python 3.10.7


SebfraOslo:
Hi,  

I am trying to install SfePy on Python 3.10.7 (with pip) but it doesn't work. It gives me the message error:
" ERROR: Failed building wheel for sfepy
Failed to build sfepy
ERROR: Could not build wheels for sfepy, which is required to install pyproject.toml-based projects"

Thank you
Sebastien

rc:
Hi Sebastien,

Thanks for the report.  The pip install is a new feature, so rough edges are to be expected. Could you run pip with `-v` option and paste the whole output here? What is your platform?

SebfraOslo:
Hi Robert,

Thanks for your reply. 
My Platform is a PC (Windows 10). When I type pip -v, it gives:
"Usage:
  C:\Users\sebas\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\python.exe -m pip <command> [options]

Commands:
  install                     Install packages.
  download                    Download packages.
  uninstall                   Uninstall packages.
  freeze                      Output installed packages in requirements format.
  inspect                     Inspect the python environment.
  list                        List installed packages.
  show                        Show information about installed packages.
  check                       Verify installed packages have compatible dependencies.
  config                      Manage local and global configuration.
  search                      Search PyPI for packages.
  cache                       Inspect and manage pip's wheel cache.
  index                       Inspect information available from package indexes.
  wheel                       Build wheels from your requirements.
  hash                        Compute hashes of package archives.
  completion                  A helper command used for command completion.
  debug                       Show information useful for debugging.
  help                        Show help for commands.

General Options:
  -h, --help                  Show help.
  --debug                     Let unhandled exceptions propagate outside the main subroutine, instead of logging them to       
                              stderr.
  --isolated                  Run pip in an isolated mode, ignoring environment variables and user configuration.
  --require-virtualenv        Allow pip to only run in a virtual environment; exit with an error otherwise.
  -v, --verbose               Give more output. Option is additive, and can be used up to 3 times.
  -V, --version               Show version and exit.
  -q, --quiet                 Give less output. Option is additive, and can be used up to 3 times (corresponding to WARNING,   
                              ERROR, and CRITICAL logging levels).
  --log <path>                Path to a verbose appending log.
  --no-input                  Disable prompting for input.
[...]

PR #934, opened 2023-02-07
Fix build dependency for Python 3.9

yosefm:
As mentioned in: https://github.com/sfepy/sfepy/issues/879

Meshio has an undeclared dependency on NumPy >= 1.20 (due to use of NumPy typings). They use both a setup.cfg file and a pyproject.toml, one of which doesn't mention the version constraints, and the duplication results in wheels depending on numpy without version constraint.

SFEPy build requirements specify `oldest-supported-numpy`, which for Python 3.9 is 1.19.3. This means that when pip tries to build in an isolated environment, it installs 1.19.3 and then fails when importing meshio at setup.py run time. This hinders use of SFEPy in automeated enviroments such as CI or poetry, when using Python 3.9.

Even if meshio was fixed to specify the constraint in the wheels, it will not solve the problem, but move it to dependency resolution time.

My proposed fix: specify the requirement in SFEPy build dependencies.

Note that this is unrelated to the NumPy that is installed as *runtime* requirement for SFEPy - it's just the one chosen for the pip build environment.

### validation
By inserting a print of the numpy version in code that runs at setup.py time, and running in different conda environments having different Python versions, I get:

| Python | NumPy |
|---------|----------|
|     3.9   |    1.20.3 |
|     3.10 |    1.21.6 |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,8 @@ requires = [
 
     "matplotlib",
     "meshio",
-    "oldest-supported-numpy",
+    "numpy==1.20.*; python_version <= '3.9'", # meshio undeclared constraint.
+    "oldest-supported-numpy; python_version > '3.9'",
     "pyparsing",
     "scipy",
     "tables",


rc:
Thanks! For clarification, what command do you use in the conda environments?

yosefm:
> Thanks! For clarification, what command do you use in the conda environments?

`pip install sfepy`

yosefm:
> Thanks! For clarification, what command do you use in the conda environments?

In some other project:
`pip install ../sfepy`
To use my local version of the source.

rc:
OK, thanks. Let us merge it!

PR #935, opened 2023-02-08
Update mirror regions

vlukes:
The final aim is to mirror different entities, e.g. surfece of volume (tdim=3) and volume (tdim=2).


vlukes:
Yes, it is finished for now. The continuation will be in the next PRs.

PR #937, opened 2023-02-08
Mappings refactoring

vlukes:
This PR removes `refmaps.c`, `mappings.pyx,` and `mappings.pxd`. `CMapping` from `mappings.pyx` is replaced by class `common/mappings.DMappings` and  by function  `fen/mappings.eval_mapping_data_in_qp()`. The interface to term C functions is realized by `DMappings.cmap` which is C data container defined in `cmapping.pyx`/`cmapping.pxd` and `refmaps.h`.


vlukes:
I will add some docstrings.

sfepy/discrete/fem/mappings.py
```diff
@@ -71,84 +193,85 @@ def get_physical_qps(self, qp_coors):
...
 
+        Parameters
+        ----------
+        qp_coors: numpy.ndarray
+            The coordinates of the integration points.
+        weights:
+            The integration weights.
+        poly_space: PolySpace instance
+            The PolySpace instance.
+        ori: ?
```
vlukes, line 210: @rc, please, could you explain `ori` and `transform`?

sfepy/discrete/common/mappings.py
```diff
@@ -4,6 +4,41 @@
...
+    Data for C functions translated to FMFields and embedded in CMapping.
+    """
+    def __init__(self, bf, det, volume, bfg, normal, dim):
+        self.bf = bf
+        self.det = det
+        self.volume = volume
+        self.bfg = bfg
+        self.normal = normal
+
+        self.cmap = CMapping(bf, det, volume, bfg, normal, dim)
```
rc, line 22: So PyCMapping instance is called geo, and we have geo.cmap.geo. Food for thought... maybe something clearer might be devised. The lowest level cmap.geo is dictated by the Cython/C terms code. What about the upper geo? :)

sfepy/discrete/fem/mappings.py
```diff
@@ -5,11 +5,122 @@
 
 from sfepy import Config
 from sfepy.base.base import get_default, output
-from sfepy.base.mem_usage import raise_if_too_large
```
rc, line 8: I would like to preserve this. It allows an early failure when memory will not be sufficient for term evaluations -> infinite swapping.

sfepy/discrete/fem/mappings.py
```diff
@@ -5,11 +5,122 @@
...
+    conn: numpy.ndarray
+        The element connectivity.
+    dim: int
+        The space dimension.
+    n_ep: int
+        The number of element points.
+    bf_g: numpy.ndarray
+        The derivatives of the domain base functions with respect to the
+        reference coordinates.
+    weight: numpy.ndarray
```
rc, line 31: FWIW: While both coor and coors are used, this is weights in Python code.

sfepy/discrete/fem/mappings.py
```diff
@@ -5,11 +5,122 @@
...
+        The weights of the integration points.
```
rc, line 32: I would use quadrature instead of integration.

sfepy/discrete/fem/mappings.py
```diff
@@ -71,84 +192,85 @@ def get_physical_qps(self, qp_coors):
...
+              - the boundary connectivity
+              - the derivatives of the domain boundary base functions with
+                respect to the reference coordinates
+
         Returns
         -------
-        cmap : CMapping instance
-            The volume mapping.
+        pycmap: PyCMapping instance
+            The domain mapping fields.
```
rc, line 226: 'The domain mapping data.' to avoid mistaking it with fields?

sfepy/terms/terms_biot.py
```diff
@@ -150,7 +150,8 @@ def function(out, val_qp, mat, vg, fmode):
             status = 0
 
         else:
-            status = terms.de_cauchy_stress(out, val_qp, mat, vg, fmode)
+
+            status = terms.de_cauchy_stress(out, val_qp, mat, vg.cmap, fmode)
```
rc, line 154: Cannot Term.get_mapping() return directly vg.cmap as before this PR?

rc:
Thanks, well done! The tests pass but I have a suspect - hanging nodes:

```
python3 sfepy/examples/diffusion/laplace_refine_interactive.py output
...
sfepy: nls: iter: 0, residual: 7.827758e+05 (rel: 1.000000e+00)
sfepy:   residual:    0.38 [s]
sfepy:     matrix:    0.00 [s]
sfepy:      solve:    0.00 [s]
sfepy: warning: linear system solution precision is lower
sfepy: then the value set in solver options! (err = 3.807985e-10 < 1.000000e-10)
sfepy: nls: iter: 1, residual: 4.133822e-10 (rel: 5.280978e-16)
sfepy: solved in 1 steps in 0.39 seconds
```
while the original code leads to:
```
sfepy: nls: iter: 0, residual: 4.785582e+00 (rel: 1.000000e+00)
sfepy:   residual:    0.01 [s]
sfepy:     matrix:    0.00 [s]
sfepy:      solve:    0.00 [s]
sfepy: nls: iter: 1, residual: 2.571117e-15 (rel: 5.372632e-16)
sfepy: solved in 1 steps in 0.01 seconds
```
Notice the residual times and the initial residual magnitudes.

rc:
The initial residual calculation is much slower now (see e.g. the navier-stokes example) - can this be improved?

sfepy/discrete/fem/mappings.py
vlukes: OK, I thought it was only related to allocation i C. And now nothing is allocated in C.

sfepy/discrete/fem/mappings.py
vlukes: Yes, `mapping data` it is better.

sfepy/terms/terms_biot.py
vlukes: The Python structure is argument of the Python terms, the C structure enters the C terms.  And, in general, I can not determine which function will be called at the time `Term.get_mapping()` is called.

rc: OK, thanks.

vlukes:
> Notice the residual times and the initial residual magnitudes.

Fixed!

vlukes:
> The initial residual calculation is much slower now (see e.g. the navier-stokes example) - can this be improved?

Yes, this can be improved using `invs_fast`.

rc:
Thanks! Even with the `*_fast()` functions, it is about 0.2 vs. 0.03 with `sfepy-run sfepy/examples/navier_stokes/navier_stokes.py`. I wonder why the old code was so fast. :)

vlukes:
> Thanks, well done! The tests pass but I have a suspect - hanging nodes:
> 
> ```
> python3 sfepy/examples/diffusion/laplace_refine_interactive.py output
> ...
> sfepy: nls: iter: 0, residual: 7.827758e+05 (rel: 1.000000e+00)
> sfepy:   residual:    0.38 [s]
> sfepy:     matrix:    0.00 [s]
> sfepy:      solve:    0.00 [s]
> sfepy: warning: linear system solution precision is lower
> sfepy: then the value set in solver options! (err = 3.807985e-10 < 1.000000e-10)
> sfepy: nls: iter: 1, residual: 4.133822e-10 (rel: 5.280978e-16)
> sfepy: solved in 1 steps in 0.39 seconds
> ```
> 
> while the original code leads to:
> 
> ```
> sfepy: nls: iter: 0, residual: 4.785582e+00 (rel: 1.000000e+00)
> sfepy:   residual:    0.01 [s]
> sfepy:     matrix:    0.00 [s]
> sfepy:      solve:    0.00 [s]
> sfepy: nls: iter: 1, residual: 2.571117e-15 (rel: 5.372632e-16)
> sfepy: solved in 1 steps in 0.01 seconds
> ```
> 
> Notice the residual times and the initial residual magnitudes.

This is fixed by [fix 2x2 inversion](https://github.com/sfepy/sfepy/pull/937/commits/755c10998f2200f508c3f5e96fbea1ffbf709ce5).

rc:
What about replacing the Python `dets_fast()` by a Cython version, that could be as fast as the original `geme_det3x3()` C function? Maybe another PR?

vlukes:
> What about replacing the Python `dets_fast()` by a Cython version, that could be as fast as the original `geme_det3x3()` C function? Maybe another PR? Actually are we sure that is the sole culprit of the slow-down? I will make some measurements when time permits.

According to my measurement, the Python `dets_fast()` and  `invs_fast()` are fast enough. There is a higher overhead in the Python implementation of `eval_mapping_data_in_qp()` compared to the original C code.

rc:
OK, so it;s the einsum calls and/or coors expansion to connectivity. Let's keep it as it is then, we can try optimizing later, if need arises.

rc:
FWIW and for the record: The `python3 sfepy/examples/diffusion/laplace_refine_interactive.py output` now works, but I have noticed earlier (*) and this confirms the observation that einsum-based evaluations have different round-off error behavior than our original C functions:
- new (einsum):
```
sfepy: nls: iter: 0, residual: 4.785582e+00 (rel: 1.000000e+00)
sfepy:   residual:    0.01 [s]
sfepy:     matrix:    0.00 [s]
sfepy:      solve:    0.00 [s]
sfepy: nls: iter: 1, residual: 2.746520e-15 (rel: 5.739156e-16)
```
- original (C):
```
sfepy: nls: iter: 0, residual: 4.785582e+00 (rel: 1.000000e+00)
sfepy:   residual:    0.01 [s]
sfepy:     matrix:    0.00 [s]
sfepy:      solve:    0.00 [s]
sfepy: nls: iter: 1, residual: 2.571117e-15 (rel: 5.372632e-16)
```
Let us keep that in mind.

(*) E.g. the symmetric `de_*` based matrices are slightly less machine-precision-symmetric than the original C ones.

Anyway, this PR is a step in the right direction, let us merge it.

PR #945, opened 2023-02-20
Explain the relations between variables in example

yosefm:
Elastodynamic problem defines 3 "independent" variables that are made dependent only by solver magic, instead of explicitly defining `v = du/dt; a = dv/dt`. This was a real head-scratcher for me, and to the best of my knowledge not documented anywhere.

### DIFF
--- a/sfepy/examples/linear_elasticity/elastodynamic.py
+++ b/sfepy/examples/linear_elasticity/elastodynamic.py
@@ -215,6 +215,12 @@ def define(
         'i' : 2,
     }
 
+    # Notes:
+    # 1. The order of the variables in the solution vector is specified here (3rd tuple member),
+    #    since that specific order is expected by the elastodynamic time-stepping solvers.
+    # 2. For the same reason, we won't explicitly define below the equations du = du/dt
+    #    and ddu = ddu/dt - these are implicitly defined by the time-stepping solver.
+    #    see the `step()` method of the solvers.
     variables = {
         'u' : ('unknown field', 'displacement', 0),
         'du' : ('unknown field', 'displacement', 1),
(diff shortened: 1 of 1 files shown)


rc:
Hi @yosefm ,

this is explained in the examples docstring under the Notes heading. The solution with the three variables is not ideal, but it was a simple way of getting the individual matrices as required by the solvers. Defining  du = du/dt and ddu = ddu/dt would not help as then there had been no way of automatically extracting the "M", "C" and "K" blocks. The alternative was to explicitly encode the time-stepping scheme into the equations which I wanted to avoid.

I would like to extend the solvers for multi-physical simulations (e.g. piezo-elasticity), so there is going to be a significant overhauling.

yosefm:
This was not to say that the choice is wrong. Just to try to explain better. I did read the Notes section before suggesting this change. On more careful reading it does say that the solver expects 3 blocks, but don't elaborate about specifying their order or about the implicit temporal equations. It does refer the reader to the 'variables' section, that's why I saw this as the fit place to add more explanation.

sfepy/examples/linear_elasticity/elastodynamic.py
```diff
@@ -223,6 +223,12 @@ def post_process(out, problem, state, extend=False):
         'dv' : ('test field', 'displacement', 'du'),
         'ddv' : ('test field', 'displacement', 'ddu'),
     }
+    # Notes:
+    # 1. The order of the variables in the solution vector is specified here (3rd tuple member),
+    #    since that specific order is expected by the elastodynamic time-stepping solvers.
+    # 2. For the same reason, we won't explicitly define below the equations du = du/dt
+    #    and ddu = ddu/dt - these are implicitly defined by the time-stepping solver.
+    #    see the `step()` method on the solvers.
```
rc, line 231: ```see the `step()` method on the solvers``` -> ```see the `step()` method of the solvers.```

rc:
I see OK. What about putting the added comment note above the variables?

rc:
Thanks, LGTM!

PR #949, opened 2023-02-22
de_grad term: grad can be multiplied by vector or matrix

vlukes:
This PR allows to multiply the gradient by either a vector or a matrix, not only by a scalar.

### DIFF
--- a/sfepy/terms/terms_multilinear.py
+++ b/sfepy/terms/terms_multilinear.py
@@ -1497,7 +1497,9 @@ class EGradTerm(ETermBase):
 
     .. math::
         \int_{\Omega} \nabla \ul{v} \mbox { , }
-        \int_{\Omega} c \nabla \ul{v}
+        \int_{\Omega} c \nabla \ul{v} \mbox { , }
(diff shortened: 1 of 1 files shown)


sfepy/terms/terms_multilinear.py
```diff
@@ -1516,8 +1519,16 @@ def get_function(self, mat, virtual, mode=None, term_mode=None,
...
+            if mat.shape[-2:] == (1, 1):
+                mul = '00'
+            elif mat.shape[-2:] in [(dim, 1), (dim, dim)]:
+                mul = 'jk'
+            else:
+                raise ValueError(f'wrong matrial shape! ({mat.shape[-2:]})')
+
             fun = self.make_function(
-                '00,i.j', mat, virtual, diff_var=diff_var,
+                mul + ',i.j', mat, virtual, diff_var=diff_var,
```
rc, line 1531: It might be more readable to either put the entire `make_function()` call into the `if` branches (like other terms do), or at least use something like `expr = '00,i.j'`, ...

rc:
Thanks, LGTM!

PR #954, opened 2023-02-28
Mappings refactoring II.

vlukes:
`SurfaceMapping` and `VolumeMapping` are replaced by `FEMapping`.

### DIFF
--- a/sfepy/discrete/fem/fields_base.py
+++ b/sfepy/discrete/fem/fields_base.py
@@ -23,7 +23,7 @@ from sfepy.discrete.fem.mesh import Mesh
 from sfepy.discrete.fem.meshio import convert_complex_output
 from sfepy.discrete.fem.utils import (extend_cell_data, prepare_remap,
                                       invert_remap, get_min_value)
-from sfepy.discrete.fem.mappings import VolumeMapping, SurfaceMapping
+from sfepy.discrete.fem.mappings import FEMapping
 from sfepy.discrete.fem.fe_surface import FESurface, FEPhantomSurface
 from sfepy.discrete.integrals import Integral
 from sfepy.discrete.fem.linearizer import (get_eval_dofs, get_eval_coors,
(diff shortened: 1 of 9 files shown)


sfepy/discrete/fem/fields_base.py
```diff
@@ -1061,19 +1059,21 @@ def create_mapping(self, region, integral, integration,
                 else:
                     se_bf_bg, se_ebf_bg, se_conn = None, None, None
 
-                sg = mapping.get_mapping(qp.vals[0], qp.weights,
-                                         poly_space=Struct(n_nod=bf.shape[-1]),
-                                         extra=(se_conn, se_bf_bg, se_ebf_bg))
+                n_ep = bf.shape[-1]
+                out = mapping.get_mapping(qp.vals[0], qp.weights,
+                                          poly_space=Struct(n_nod=n_ep),
```
rc, line 1064: I wonder if, instead of the fake poly_space, something better could be used - either a true PolySpace instance or None. Maybe None, because `n_ep == se_conn.shape[1]`?

rc:
Thanks! LGTM so far.

sfepy/discrete/fem/fields_base.py
vlukes: This is solved by the next commit.

sfepy/discrete/fem/mappings.py
```diff
@@ -209,6 +203,8 @@ def get_mapping(self, qp_coors, weights, poly_space=None, ori=None,
             The coordinates of the integration points.
         weights:
             The integration weights.
+        bf: numpy.ndarray
+            The base functions.
```
rc, line 207: We should use 'basis function' instead of 'base function'. Mea culpa...

rc:
OK, thanks!

sfepy/discrete/fem/mappings.py
vlukes: I agree.

PR #956, opened 2023-03-01
fix material by region for mixed cell/facet regions

### LINKED ISSUE #955: per-region material definition: using single Material in both volume and surface integrals does not work
See [this ML thread](https://mail.python.org/archives/list/sfepy@python.org/thread/AFS6RE54ZMREVYDDP3GSAEJREZVPEVTK/#6LPSEXO3B6ANIXD3ED5IMOIJZJIQWJ2O).

rc:
Fixes #955.


vlukes (approved):
:+1:

PR #958, opened 2023-03-01
Unification of fe fields

vlukes:
FE fields `VolumeField` and `SurfaceField` are unified in `FEField`.

### DIFF
--- a/sfepy/discrete/fem/fields_base.py
+++ b/sfepy/discrete/fem/fields_base.py
@@ -236,10 +236,6 @@ class FEField(Field):
         Assumes one cell type for the whole region!
         """
         shape = parse_shape(shape, region.domain.shape.dim)
-        if not self._check_region(region):
-            raise ValueError('unsuitable region for field %s! (%s)' %
-                             (name, region.name))
-
         Struct.__init__(self, name=name, dtype=dtype, shape=shape,
                         region=region)
         self.domain = self.region.domain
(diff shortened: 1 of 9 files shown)


sfepy/discrete/fem/fields_base.py
```diff
@@ -236,10 +242,6 @@ def __init__(self, name, dtype, shape, region, approx_order=1):
         Assumes one cell type for the whole region!
         """
         shape = parse_shape(shape, region.domain.shape.dim)
-        if not self._check_region(region):
-            raise ValueError('unsuitable region for field %s! (%s)' %
-                             (name, region.name))
-
         Struct.__init__(self, name=name, dtype=dtype, shape=shape,
```
rc, line 245: Does it mean all regions are now usable? Or just cell and facet ones?

sfepy/discrete/fem/fields_base.py
```diff
@@ -629,7 +702,8 @@ def restore_substituted(self, vec):
         vec = vec.reshape((self.n_nod, self.n_components)).copy()
         evec = vec[self.econn]
 
-        vec[self.econn0] = nm.einsum('cji,cjk->cik', self.basis_transform, evec)
+        vec[self.econn0] = nm.einsum('cji,cjk->cik', self.basis_transform,
+                                     evec)
```
rc, line 706: Maybe turn on optimization too? But for two operands it IMO does not matter.

rc:
Thanks!

I noticed that the volume->cell, surface->facet change needs to be done also in the docs.

sfepy/discrete/fem/fields_base.py
```diff
@@ -1438,42 +1257,53 @@ def average_qp_to_vertices(self, data_qp, integral):
...
 
         return data_vertex
 
+    # from SurfaceField()
+    # def _setup_bubble_dofs(self):
+    #     """
+    #     Setup bubble DOF connectivity.
+    #     """
+    #     return 0, None, None
+
```
rc, line 1309: SurfaceField was defined on faces of a volume mesh, so it did not have bubble DOFs, only vertex, edge and face. How is this solved now?

sfepy/discrete/fem/fields_base.py
vlukes: `optimization=True` costs nothing and does no harm

sfepy/discrete/fem/fields_base.py
vlukes: Yes, just cell and facet regions are usable. Fixed in `._setup_geometry()`.

sfepy/discrete/fem/fields_base.py
vlukes: What should return this function for `VolumeField` (`self.is_surface` is `False`)?

rc: See e.g. https://github.com/sfepy/sfepy/blob/master/sfepy/discrete/fem/fields_nodal.py

rc:
LGTM from my side, thanks. Let me know when it is ready. You plan to update the docs, right?

rc:
I tried to update the docs, please check when you have time.

rc:
So I guess it is done now. I will merge soon.

vlukes:
> So I guess it is done now. I will merge soon.

Do it.

Issue #944, opened 2023-02-18
Modernise build system


yosefm:
The current extension modules (and general setup process) are handled by numpy.distutils, which is deprecated and will eventually have to go. Until it goes, there is some danger of a new setuptools version braking things. As the NumPy project [suggests](https://numpy.org/doc/stable/reference/distutils_status_migration.html#numpy-setuptools-interaction), we should consider two alternatived: meson (used by SciPy) and scikit-build.

After some research I think that scikit-build has more potential:
1. More of a drop-in replacement in the setup.py region. 
2. meson-python module documentation warns that windows/OSX support is incomplete. That might be out of date, as SciPy uses it, but I don't know.
3. As the scikit-build creator [points out](https://iscinumpy.gitlab.io/post/scikit-build-proposal/), it uses CMake, which is more widely used in scientific code that meson, and therefore gives more compatibility with other code. 

although with experience I might change my mind :) 

That said, I've only started to come up with an example. [The branch](https://github.com/yosefm/sfepy/tree/numpy-distutils-byebye) creates a parallel `setup_test.py` which only builds a couple of shared libraries (the rest are not built) in `sfepy.discrete.common.extmods`, just to get a feel for how it may look. `CMakeLists.txt` files replace the `setup.py` files, and subpackages are supported out of the box by CMake.

PR #960, opened 2023-03-04
replace Numpy distutils with scikit-build

yosefm:
In response to issue #944 - this draft PR adds `setup_test.py` for building sfepy with scikit-build. As we move from the draft phase I imagine this would become simply `setup.py` and all the previous system will be removed, but for now I'm using it to compare builds.

What it does: builds all extensions and installs all the package. Passes the test suite on Linux. 

What it doesn't do:
- not tested on windows yet.
- does not consider `site_cfg.py`. 
  - Some of the things there can be directly handled by CMake. For example, we can set a build type or directly set cmake flags in `CMakeLists.txt` or in [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html). we don't need to configure system type (windows/posix) since CMake has facilities for that.
  - I do wish to get feedback on what cflags to use. Currently they are determined by `site_cfg.py`, which means that building on Windows might get the wrong flags. I don't understand this completely because I did build on Windows in the past without having to change site_cfg. Anyway we should consider using CMake for that. CMAke has default values based on build type (-O3 for release) which may be different than the current site_cfg. Also, `numpy.distutils` seems to add stack protection and stuff, which I will have to replicate and test on both windows and Linux.
  - debug flags can also be set directly in CMakeLists.txt, or else there can be a generated include file for cmake to define them from site_cfg, but this feels like overkill to me.
- not all `setup.py` features have been replicated (yet), particularly:
  - checking and warning on missing dependencies. This will return when both setups are merged later.
  - examples not installed. I added our own reduced facility to add a whole data dir (see `setup_test.py`, to replace the facility found in `numpy.distutils`, but perhaps a better option would be to move data dirs into the package as [promoted by recent setuptools](https://setuptools.pypa.io/en/latest/userguide/datafiles.html).

Notes:
- Some extensions compiled against individual files from `discrete/common/extmods` are now just linked against `sfepy_common`
- The final PR will actually reduce code size due to removal of all subpackage setup files
In summary, there is some finishing work to do, but first there are some decisions to make. I open this PR as draft to get the discussion going before finalization.


sfepy/discrete/common/extmods/_geommech.pyx
```diff
@@ -7,8 +7,7 @@ cimport cython
 cimport sfepy.discrete.common.extmods._geommech as gmch
 from types cimport int32
 
-from sfepy.discrete.common.extmods._fmfield cimport\
-    (FMField, FMF_SetCell, array2fmfield4)
+from sfepy.discrete.common.extmods._fmfield cimport (FMField, FMF_SetCell, array2fmfield4)
```
yosefm, line 10: Due to an unhadled issue in skbuild, https://github.com/scikit-build/scikit-build/issues/302

rc:
I only briefly glanced over the build configuration files, but it looks simple and readable!

> In response to issue #944 - this draft PR adds `setup_test.py` for building sfepy with scikit-build. As we move from the draft phase I imagine this would become simply `setup.py` and all the previous system will be removed, but for now I'm using it to compare builds.

The github tests pass, so it looks working also on macOS and windows.
 
> What it does: builds all extensions and installs all the package. Passes the test suite on Linux.
> 
> What it doesn't do:
> 
>     * not tested on windows yet.
> 
>     * does not consider `site_cfg.py`.
>       
>       * Some of the things there can be directly handled by CMake. For example, we can set a build type or directly set cmake flags in `CMakeLists.txt` or in [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html). we don't need to configure system type (windows/posix) since CMake has facilities for that.

The `site_cfg.py` file was inspired by the SciPy many years ago. It serves there to "provide configuration information about non-Python dependencies", which we have too, but use them through other packages (e.g. BLAS of NumPy etc). So if the file is found superfluous, I am not against replacing/removing it.

>       * I do wish to get feedback on what cflags to use. Currently they are determined by `site_cfg.py`, which means that building on Windows might get the wrong flags. I don't understand this completely because I did build on Windows in the past without having to change site_cfg. Anyway we should consider using CMake for that. CMAke has default values based on build type (-O3 for release) which may be different than the current site_cfg. Also, `numpy.distutils` seems to add stack protection and stuff, which I will have to replicate and test on both windows and Linux.
>
>       * debug flags can also be set directly in CMakeLists.txt, or else there can be a generated include file for cmake to define them from site_cfg, but this feels like overkill to me.

I am mostly using just the `is_release` flag when making releases. Then being able to pass the debug flags 
`-DDEBUG_FMF`, `-DDEBUG_MESH` is also needed. Then `refmap_memory_factor` is used in Python code. So those are the things I would like to be able to configure somehow. The other options can either be delegated to CMake or just removed (e.g. `python_version` is a remnant from Python 2/3 transition).

>     * not all `setup.py` features have been replicated (yet), particularly:
>       
>       * checking and warning on missing dependencies. This will return when both setups are merged later.

FWIW, the package versions listed in `sfepy/version.py` are not very relevant and need updating. The current checking and warning on missing dependencie has also some flaws like one cannot build the extension modules without e.g. meshio installed, although it is not a build-time dependency.

>       * examples not installed. I added our own reduced facility to add a whole data dir (see `setup_test.py`, to replace the facility found in `numpy.distutils`, but perhaps a better option would be to move data dirs into the package as [promoted by recent setuptools](https://setuptools.pypa.io/en/latest/userguide/datafiles.html).

Or we will just make them to be part of the package?  The same goes for the tests. The ability to run tests for the installed package needs to be preserved.

> Notes:
> 
>     * Some extensions compiled against individual files from `discrete/common/extmods` are now just linked against `sfepy_common`

OK. The `sfepy_common` is not very big, so if that makes things simpler...
 
>     * The final PR will actually reduce code size due to removal of all subpackage setup files
>       In summary, there is some finishing work to do, but first there are some decisions to make. I open this PR as draft to get the discussion going before finalization.

Thanks!

sfepy/discrete/common/extmods/_geommech.pyx
rc: Would using
```
from sfepy.discrete.common.extmods._fmfield cimport (
    FMField, FMF_SetCell, array2fmfield4,
)
```
also help? I would avoid the line longer than 79 chars, but if it is not possible, so be it :)

rc:
Thanks @yosefm! Could you rebase the PR on top of the current master? This worked for me without conflicts:
```
git checkout numpy-distutils-byebye
git rebase --autostash master
```
I will have a look at the CI issues.

yosefm:
The difference seems to be pip. When doing `python setup.py install` it is no problem to run pytest later (through coverage or alone). Only when doing `pip install .` (editable or not) then the tests fail.

rc:
Does `pip install` build the extension modules properly?

yosefm:
Looks like it does, running with --verbose I can see the files are built and placed in the same relative location in the build temp directory (_skbuild/...)
There must be some naming thing I'm missing.

rc:
The CI output seems not to show details on the building - maybe add --verbose to the CI config? I have just tried re-running a Python 3.9 job to see whether it is Python version dependent.

rc:
Yeah, the same error.

yosefm:
If I build a wheel and install it with pip, all cool. If I let pip do the build in its own temp. environment, then the problem arises.

yosefm:
Now this works locally with pip install (not editable) but not on the CI agent. I'm not sure how to debug it.

I had to steal a hack from numpy.distutils, which means we'll have a dependency on distutils to remove later, but that's still more focused than depending on numpy.distutils.

rc:
Now it looks that the VERSION file is not generated. The lines 154-156 in `setup_orig.py` that do that:
```
    fd = open('VERSION', 'w')
    fd.write(VERSION)
    fd.close()
```
are missing in the new `setup.py`.

yosefm:
Wow, I would have hunted that a whole week :)

Ok, now we're in a good place because I know what's going on. `coverage` adds '.' to `sys.path`, the same as running `python -m pytest` instead of just `pytest`. Since this isn't an editable install, and the extmods really aren't in the current directory tree, it fails. Running just pytest on the installed tests would pass, and so would anything that uses the installed sfepy instead of the current directory. 

Exactly how it would look technically - needs to be worked out.

yosefm:
Note that the coverage run now happens in a specially created subdirectory, to avoid using the local sources instead of the installed package. Another way would have been to insist on using an "editable" install, but with skbuild that requires explicitly running `setup.py build_ext --inplace`, and thus installing build dependencies, so seemed too heavy for running tests in CI. It's still usable as a local development option.

yosefm:
> Hi! Just curious since I was following the PR reach the state where it seems to be mature: even if not using an editable install, build dependencies are all needed anyway, right? Shouldn't both ways work the same? I'm asking from the CI perspective since I use something very similar at work. Thanks!

Build dependencies will be downloaded in the pip build environment, and discarded with that environment. They will not be installed on the main environment, where the tests are run. For CI, that's about the same load as the editable install done by numpy.distutils, only it may make things cleaner by catching runtime use of build-only dependencies. I mention the whole thing mostly because there may be people using some flow based on an editable install, and they will need some adaptation.

rc:
Thanks for the updates!

I am using the editable install when developing sfepy - based on what you wrote it seems that it is supported by the new build system, but then are there any additional steps to do besides `pip install -e .`?

I have also tried to make a tentative PR https://github.com/conda-forge/sfepy-feedstock/pull/50, in anticipation of the next sfepy release due the end of March. I do not know why it fails, so any comment on that would be appreciated.

yosefm:
Editable install won't work without ` setup.py build_ext --inplace`.

rc:
I have tried to build sfepy on my linux box (an ubuntu 22.04 flavor). These are additional steps I had to do.

- I had to install scikit-build, cmake and ninja - those should be added to [requirements](http://sfepy.org/doc-devel/installation.html#requirements).

- Then I had to tweak my `site_cfg.py` to make `compile_flags` a list. Maybe allow also a string by using:
```
    def compile_flags(self):
        if has_attr(site_cfg, 'compile_flags'):
            flags = site_cfg.compile_flags
            if isinstance(flags, str):
                flags = flags.split()

        else:
            flags = ['-g', '-O2']
```
What do you think?

- Then the build process proceeded and all tests passed.

yosefm:
> I have tried to build sfepy on my linux box (an ubuntu 22.04 flavor). These are additional steps I had to do.
> 
>     * I had to install scikit-build, cmake and ninja - those should be added to [requirements](http://sfepy.org/doc-devel/installation.html#requirements).

Sure. 

>     * Then I had to tweak my `site_cfg.py` to make `compile_flags` a list. Maybe allow also a string by using:
> 
> 
> ```
>     def compile_flags(self):
>         if has_attr(site_cfg, 'compile_flags'):
>             flags = site_cfg.compile_flags
>             if isinstance(flags, str):
>                 flags = flags.split()
> 
>         else:
>             flags = ['-g', '-O2']
> ```

Could be, but also deleting site_cfg regenerates it, so consider if it's worth having another special case to maintain. If you think it's preferable, I'll add it.

rc:
> > I have tried to build sfepy on my linux box (an ubuntu 22.04 flavor). These are additional steps I had to do.
> > ```
> > * I had to install scikit-build, cmake and ninja - those should be added to [requirements](http://sfepy.org/doc-devel/installation.html#requirements).
> > ```
> 
> Sure.
> 
> > ```
> > * Then I had to tweak my `site_cfg.py` to make `compile_flags` a list. Maybe allow also a string by using:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >     def compile_flags(self):
> >         if has_attr(site_cfg, 'compile_flags'):
> >             flags = site_cfg.compile_flags
> >             if isinstance(flags, str):
> >                 flags = flags.split()
> > 
> >         else:
> >             flags = ['-g', '-O2']
> > ```
> 
> Could be, but also deleting site_cfg regenerates it, so consider if it's worth having another special case to maintain. If you think it's preferable, I'll add it.

Yes, deleting helps but then we would need to document it somewhere and prepare for people asking :). It seems easier to allow the original way and maybe print a deprecation warning?

rc:
I have found that this build system is almost three times faster than the original one, nice!

There is still a thing to do - fix `python setup.py sdist` and pip use.

It seems that the sdist command does not obey `MANIFEST.in` file. It includes some surplus files and directories, notably `_skbuild/` and `eployment-CI/`. Anyway, then I uploaded the sdist file to testpypi, tested and the tests failed.

I did:
```
# set version to any version not uploaded yet to testpypi (2023.1.2 in my case), set is_release = True in site_cfg.py
python3 setup.py sdist

python3 -m twine upload --repository-url https://test.pypi.org/legacy/ dist/sfepy-2023.1.2.tar.gz

python3 -m venv venv
source venv/bin/activate
python3 -m pip install -U -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ sfepy
python3 -m pip install pytest
sfepy-test
```
and got a bunch of errors like:
```
ModuleNotFoundError: No module named 'sfepy.discrete.common.extmods.cmapping'
```
So it looks like the extension modules were not built.

yosefm:
I have just spent an hour trying to reproduce the spurious includes. I can't do it now, but I did manage to reproduce it partly (include the _skbuild directory) for a while - don't know how. So my guess is, there's some caching or leftovers from other branches or a non-clean worktree.  I tried on a clean worktree and couldn't reproduce it at all.

The second issue (extmods not built from sdist) I can confirm. Looking into it.

yosefm:
With most recent commit, I seem to have the sdist issue solved. It's not very elegant, but I'm running into problems with skbuild when trying to understand if there's a more elegant way to do it. so I'm thinking maybe do it like this, and keep the investigation of how skbuild manifet works to some other time.

yosefm:
@rc  Strange to say, but it's possible that with proper removal of the _skbuild directory, the latest commit is also unnecessary. I tested locally, but it would be better if you test this too: just use the commit before last, but after removing the _skbuild (and maybe also dist) directory. If this works, then I'll revert the last commit.

yosefm:
Another directory to remove, if it exists in a worktree, is sfepy.egg-info. This makes for a clean manifest too. Anyway, with the most recent commit it's working.

rc:
Great, thanks!

I have just tweaked a bit the MANIFEST.in file, so that it ignores the many work directories that I have/use - see 
https://github.com/rc/sfepy/tree/numpy-distutils-byebye - the tarball's size looks fine without removing directories manually, and the installation from testpypi passes tests. 

I have looked at https://packaging.python.org/en/latest/guides/using-manifest-in/ and it seems that sfepy.egg-info cannot be 'un-included', so having it is not a problem.

So can you add the commit from my branch (or I can push it here), and rebase the branch to master? The MANIFEST.in will have to be updated as sfepy/discrete/common/extmods/refmaps.c was removed. That should be all!

yosefm:
done

rc:
Thanks again, merged!

PR #968, opened 2023-03-21
update elastodynamics solvers

rc:
Update for multiphysical problems (WIP), simplify problem description.


rc:
Hi @vlukes, what do you think about the simplified syntax of elastodynamics problems? See e.g. linear_elasticity/seismic_load.py or multi_physics/piezo_elastodynamic.py. Only the Newmark solver was updated for multiphysics - I will eventually update the other ones depending on how it will work for us.

PR #980, opened 2023-04-11
fix presolve in homogenization

rc:
Hi @vlukes, I have removed presolving from the MiniAppBase, as it is done in Problem.solve() anyway. Further optimizations are possible, e.g. trying the presolve only if a direct solver is selected.

What do you say to replacing the iterative solvers?

### DIFF
--- a/sfepy/discrete/problem.py
+++ b/sfepy/discrete/problem.py
@@ -266,7 +266,8 @@ class Problem(Struct):
 
             if conf is None:
                 self.conf = Struct(options={}, ics={},
-                                   ebcs={}, epbcs={}, lcbcs={}, materials={})
+                                   ebcs={}, epbcs={}, lcbcs={}, materials={},
+                                   solvers={})
 
         self.equations = equations
         self.fields = fields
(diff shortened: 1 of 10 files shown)


vlukes:
Instead of specifying  `'use_presolve': True`  in each example, I suggest setting `use_presolve` to `True` if `is_linear` in `coefs_base.init_solvers()`.

rc:
The iterative linear solvers do not have this option. And I would rather not modify user settings - you might want to set use_presolve to False for debugging.

vlukes:
> And I would rather not modify user settings - you might want to set use_presolve to False for debugging.

Yes, I agree. I thought to set `use_presolve` = True by default if not set by the user.

rc:
The problem is that the value is bool with default False, so we do not have a way to distinguish the two situations: a user set False, a user did not set the option. We would need to modify the direct solvers...

vlukes:
What about this in `coefs_base.init_solvers()`?

```
        if self.is_linear:
            for v in problem.conf.solvers.values():
                if v.kind.startswith('ls.'):
                    v.set_default('use_presolve', True)

        problem.set_conf_solvers(problem.conf.solvers, opts)
        problem.init_solvers()
        problem.set_linear(self.is_linear)
```

vlukes:
Yes, now I see that there is an issue with iterative solvers ...

rc:
OK, it introduces a complex magic, but we can try it. Note that, for correctors with a single linear system, using presolve and scipy_direct is actually a bit slower, as the matrix digest needs to be computed. What about the examples? I kind of like having it explicitly there.

rc:
OK, it is there.

PR #988, opened 2023-05-16
fix sfepy-view for pyvista>=0.39.0

rc:
Fixes displaying HDF5 results.


rc:
It works, let's merge it :)

PR #997, opened 2023-05-30
remove/replace distutils remains

### LINKED ISSUE #984: clean up numpy.distutils remnants
`git grep numpy.distutils` reveals that there are still some small pieces of code in `tools/build_helpers.py`.

rc:
- closes #984.

@yosefm, I have replaced or removed everything from distutils except the code in `tools/install_data.py`. Is there a way of replacing even that?

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,8 +6,7 @@ filterwarnings =
 #    error
     ignore:numpy.dtype size changed
     ignore:numpy.ufunc size changed
+    # Triggered by PRIMME.
+    ignore:.*the `scipy.sparse.linalg.interface` namespace is deprecated.*:DeprecationWarning
     ignore:.*is a deprecated alias for.*:DeprecationWarning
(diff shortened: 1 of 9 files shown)


yosefm:
I used it only to make sure that the same installation path used by numpy.distutils (e.g. with the package rather than someplace beside it), so that all tests etc. keep functioning. Options I thought of:

1. As suggested in the skbuild PR, it might be good to move data into the package and then setuptools treats it as equal to code, in a way. This seems to be the modern way. Only thing is, one needs to ensure that files like VERSION and LICENSE are handled correctly.
2. understand skbuild options better and convince it to do the same

HTH

yosefm:
Something like this for option 1.
https://setuptools.pypa.io/en/latest/userguide/datafiles.html

rc:
OK, thanks! In that case I will leave this undecided and merge the PR as is - My aim was to clean up all the warnings during the tests, which was achieved.

PR #1006, opened 2023-07-02
Fix some deprecations on NumPy 1.25 and newer setuptools

yosefm:
* replace `alltrue` with `all`
* replace `pkg_resources` with `packaging`


rc:
Thanks, LGTM!

PR #1009, opened 2023-07-19
Restrict Cython version

yosefm:
Cython 3.0 came out this week. It changed the default language level, and hence broke our build. This should be better understood, but until then this PR provides a stopgap measure.


rc:
OK, thanks!

PR #1024, opened 2023-09-11
Mumps multiple rhs

### LINKED ISSUE #1023: ls.mumps does not work with multiple right-hand sides
Demonstration:
```
import numpy as np
import scipy.sparse as sps

from sfepy.base.base import Struct
from sfepy.solvers import Solver

ls = Solver.any_from_conf(Struct(name='ls', kind='ls.mumps'))

num = 5
nc = 2

rhs = np.cumsum(np.ones((num, nc), dtype=np.float64), axis=1)

# mtx = sps.eye(num, dtype=np.float64)
mtx = sps.csr_matrix(
    np.array([[3, 1, 0, 0, 0],
              [1, 3, 1, 0, 0],
              [0, 1, 3, 1, 0],
              [0, 0, 1, 3, 1],
              [0, 0, 0, 1, 3],]), dtype=np.float64)

sol1 = ls(rhs, mtx=mtx)
val1 = sol1.sum(axis=0)

sol2 = []
for ic in range(nc):
    sol2.append(ls(rhs[:, ic], mtx=mtx))
sol2 = np.array(sol2).T
val2 = sol2.sum(axis=0)

print(rhs)

print(sol1)
print(sol2)

print(val1)
print(val2)
```
leads to
```
[[1. 2.]
 [1. 2.]
 [1. 2.]
 [1. 2.]
 [1. 2.]]
[[ 0.11111111  0.66666667]
 [-0.11111111  0.66666667]
 [ 0.11111111  2.        ]
 [ 1.          2.        ]
 [ 1.          2.        ]]
[[0.27777778 0.55555556]
 [0.16666667 0.33333333]
 [0.22222222 0.44444444]
 [0.16666667 0.33333333]
 [0.27777778 0.55555556]]
[2.11111111 7.33333333]
[1.11111111 2.22222222]
```
and no error is reported .

vlukes:
This should fix #1023.

### DIFF
--- a/sfepy/solvers/ls.py
+++ b/sfepy/solvers/ls.py
@@ -972,7 +972,12 @@ class SchurMumps(MUMPSSolver):
             schur_list.append(nm.arange(slc.start, slc.stop, slc.step, dtype='i'))
 
         self.mumps_ls.set_mtx_centralized(mtx)
-        out = rhs.copy()
+
+        if len(rhs.shape) == 2 and rhs.shape[1] > 1:
+            out = nm.asfortranarray(rhs)
+        else:
+            out = rhs.copy()
+
         self.mumps_ls.set_rhs(out)
 
         S, y2 = self.mumps_ls.get_schur(nm.hstack(schur_list))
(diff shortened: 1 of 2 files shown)


rc:
`nm.asfortranarray()` does not copy if the original array is already a fortran array - is that OK?

vlukes:
> `nm.asfortranarray()` does not copy if the original array is already a fortran array - is that OK?

I do not know :) But, yes, it would be safer always copy it.

PR #1032, opened 2023-10-11
spectral element method

rc:


vlukes:
:+1: 

Do you have an example for this?

rc:
I have a script that plots convergence of eigenvalue problems for the Laplace and elasticity matrices and for different bases. Where to put it? I am thinking about `sfepy/examples/miscellaneous/`. We should somehow include the non-declarative examples to the gallery, so that those are more visible.

rc:
@vlukes I have added the example, what do you think? I have also a working solution for having the interactive examples in gallery - it will be in my next PR.

vlukes:
> Where to put it? I am thinking about `sfepy/examples/miscellaneous/`

I think that `sfepy/examples/miscellaneous/` is a good place for such examples.

PR #1049, opened 2023-12-07
better eigenvalue solver for modal_analysis_declarative.py

flothesof:
Hi!
I’ve noticed you added this nice modal analysis example.
I think the solver choice is not optimal for two reasons:
- primme is not builtin with scipy 
- most importantly, when using the eigsh solver for finding the lowest frequency mode, it is important to use the shift-invert sigma parameter set to 0 and to search for the largest magnitude eigenmodes (LM), something that took me a long time to find out and is essential when you want the solve to be fast. This is not documented anywhere else AFAIK and would thus be interesting for users of this example.

Hence my proposed changes.

Regards,
Florian

### DIFF
--- a/sfepy/examples/linear_elasticity/modal_analysis_declarative.py
+++ b/sfepy/examples/linear_elasticity/modal_analysis_declarative.py
@@ -15,7 +15,7 @@ wheelset.vtk. View the results using::
 
 The first six frequencies calculated by SfePy::
 
-  [11.287, 11.317, 34.432, 80.709, 80.901, 93.144]
+  [11.272, 11.322, 34.432, 80.711, 80.895, 93.149]
 
 The results of modal analysis performed in Ansys::
 
--- a/sfepy/examples/linear_elasticity/modal_analysis_declarative.py
+++ b/sfepy/examples/linear_elasticity/modal_analysis_declarative.py
@@ -98,9 +98,16 @@ def define(n_eigs=6, approx_order=1, density=7850., young=210e9, poisson=0.3):
         #     'which': 'sm',
         #     'eps': 1e-6,
         # }),
-        'eig': ('eig.primme', {
-            'which': 'SM',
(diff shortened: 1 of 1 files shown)


vlukes:
Hi Florian,

> * most importantly, when using the eigsh solver for finding the lowest frequency mode, it is important to use the shift-invert sigma parameter set to 0 and to search for the largest magnitude eigenmodes (LM), something that took me a long time to find out and is essential when you want the solve to be fast. This is not documented anywhere else AFAIK and would thus be interesting for users of this example.

I was not able to get eighs to solve the lowest frequency eigenvalue problems fast, so I decided to use primme which gives reliable results in reasonable time. But your solution is really fast, much faster than primme. Thank you very much for that! 

Regards,
Vladimir

rc:
Hi @flothesof,

I had been aware of the shift-invert mode possibility (see e.g. [this](http://sfepy.org/doc-devel/examples/quantum-quantum_common.html) or [this](http://sfepy.org/doc-devel/examples/miscellaneous-refine_evp.html)), but had a problem that it often missed an eigenvalue, when compared to Matlab. PRIMME usually does a good job of not missing eigenvalues. But this is a small example, and eigsh seems to work reasonably well, so +1.

Can you update the sfepy-calculated frequencies in the docstring to those calculated by eigsh? Thank you!

rc:
> ’ve been using the sigma=0 option regularly for the last couple months for solving elastic vibration problems and haven’t had any issues. I see that the quantum example focuses on an eigenvalue near a nonzero value. Maybe in that case the routine is less reliable?

The quantum examples included in sfepy are small, so there it is OK. But yes, it might be related to the non-zero shift. I had issues with missing eigenvalues for large 3D problems often in complex numbers in various contexts.

Thanks for the update! I will merge now, ignoring the CI failure, as your first commit tested OK and now only the docstring changed.

PR #1054, opened 2023-12-19
Fix tangent modulus of Ogden model

heczis:
Hi, I finally fixed the Ogden model so that it at least gives correct results [hopefully].
It turned out that the formula that I used was not complete, so I had to use the one in Crisfield, vol. 2.
I tested it under different deformations (uniaxial, shear, isochoric, volumetric, different combinations) and it gives essentially the same results to neo-Hooke (with alpha=2) and both agree well with finite differences.

A faster implementation is next on the list.

### DIFF
--- a/sfepy/terms/terms_hyperelastic_tl.py
+++ b/sfepy/terms/terms_hyperelastic_tl.py
@@ -214,6 +214,8 @@ class OgdenTLTerm(HyperElasticTLBase):
     and :math:`\mathbf{N}^{(k)}`, :math:`k=1, 2, 3` are the eigenvectors of
     :math:`\mathbf{C}`.
 
+    Tangent stiffness is implemented according to sec. 13.8.2 in [3].
+
     :Definition:
 
     .. math::
@@ -234,6 +236,9 @@ class OgdenTLTerm(HyperElasticTLBase):
     Treloar's data. Archive of Applied Mechanics, Vol. 82, No. 9, Pp. 1183-1217
     (2012), DOI `10.1007/s00419-012-0610-z
     <https://dx.doi.org/10.1007/s00419-012-0610-z>`_.
+
+    [3] Crisfield, M. A. Non-linear Finite Element Analysis of Solids and
+    Structures. Volume 2: Advanced Topics. John Wiley and Sons (1997).
     """
     name = 'dw_tl_he_ogden'
     family_data_names = ['det_f', 'sym_c', 'tr_c', 'sym_inv_c']
@@ -253,14 +258,14 @@ class OgdenTLTerm(HyperElasticTLBase):
         return out
 
     def stress_function(self, out, mat, *fargs, **kwargs):
-        det_f, sym_c, _, _ = fargs
+        det_f, sym_c = fargs[:2]
         mat = HyperElasticBase.tile_mat(mat, det_f.shape[0])
         coef, exp = mat[:, :, :, :1], mat[:, :, :, 1:]
 
         # compute principal stretches and directions
         c_mats = sym_c[:, :, [[0, 3, 4], [3, 1, 5], [4, 5, 2]], 0]
         lbds, nks = nm.linalg.eigh(c_mats)
-        lbds = lbds**.5
+        lbds = nm.sqrt(det_f[..., 0]**(-2.0/3.0) * lbds)
 
         # evaluate stress
         n_cells, n_qps, _, _ = out.shape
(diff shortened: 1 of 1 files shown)


rc:
Thanks! The code looks smaller than the previous version :)

A few points:
- Is it the same algorithm as in Box 11.3 (p. 387) of [1], as discussed on page 385 there?
- Can you add (some of?) your testing code as an example, so that the term is tested?
- When tested with `m = Material('m', mu=14., alpha=20000., mua=(14., 20000.))` and compared with `dw_tl_he_ogden_ad` (see #1052), I get the relative error of matrices about 0.5, while the original (wrong) code
leads to about 0.2, which seems weird. For mu = 1, alpha = 2 (neo-Hookean) the relative error is 1e-12, which looks OK. Adding a test for non-neo-Hookean setting would be a good idea. Are you aware of some benchmark? I checked https://github.com/jjcremmers/PyFEM and did not find the Ogden material there.

[1] R. de Borst, M. A. Crisfield, J. J. C. Remmers, and C. V. Verhoosel, Nonlinear Finite Element Analysis of Solids and Structures, 2nd edition (Wiley, Hoboken, NJ, 2012).

heczis:
- The algorithm is not the one in [1]. I somewhat corrected the formula in [2] according to sec. 13.8.2 of [3]
- My testing code is here: https://gist.github.com/heczis/e0761930467a0243e9db335e18b45c01
  I will try to add a comparison against the 2-parameter Mooney-Rivlin tomorrow (also a special case of the Ogden model and is implemented in SfePy).
- When I set `alpha=20000`, I get `NaN`s from both the JAX-based term and the corrected one.

[1] R. de Borst, M. A. Crisfield, J. J. C. Remmers, and C. V. Verhoosel, Nonlinear Finite Element Analysis of Solids and Structures, 2nd edition (Wiley, Hoboken, NJ, 2012).
[2] Steinmann, P., Hossain, M., Possart, G. Hyperelastic models for rubber-like materials: Consistent tangent operators and suitability for Treloar's data. Archive of Applied Mechanics, Vol. 82, No. 9, Pp. 1183-1217 (2012), DOI `10.1007/s00419-012-0610-z <https://dx.doi.org/10.1007/s00419-012-0610-z>`_.
[3] Crisfield, M. A. Non-linear Finite Element Analysis of Solids and Structures. Volume 2: Advanced Topics. John Wiley and Sons (1997).

rc:
Thanks for the link to the testing code. Maybe print also relative differences, but the errors seem OK.

Can you think of an example with some nice (muscle shaped?) but small mesh? Within the example you could check the tangent matrix via the `nls.newton` option `check`.

rc:
Hi @heczis, any news?

heczis:
Sorry, all my other work went wrong in the meantime. But got to it, finally.

Here is the test (mesh and script). It seems to me that the simulation runs somehow better (but didn't test it) and even the results seem OK to me (homogeneous strain in the direction of extension).

[muscle_test.zip](https://github.com/sfepy/sfepy/files/14156315/muscle_test.zip)

However, I am not sure about the output of the checks. E.g.:

--- a-r 20 biggest (by r): ir, ic, difference, analytical, adiff, rdiff
  155   154 -1.7234e-03 -1.0376e-02  8.65e-03  5.02e+00
  154   155 -1.7234e-03 -7.5793e-04  9.65e-04  5.60e-01
   83    81  7.8685e-02  1.1980e-01  4.11e-02  5.23e-01
...

Is it good or not? How to read it?
Should the mesh have been even smaller?

rc:
Thanks for the update!

> --- a-r 20 biggest (by r): ir, ic, difference, analytical, adiff, rdiff
> 155 154 -1.7234e-03 -1.0376e-02 8.65e-03 5.02e+00
> 154 155 -1.7234e-03 -7.5793e-04 9.65e-04 5.60e-01
> 83 81 7.8685e-02 1.1980e-01 4.11e-02 5.23e-01
> ...
> Is it good or not? How to read it? For example -1.7234e-03 and -1.0376e-02 should be very close, which they are not.

The difference seems substantial.

When I replace the Ogden term with a neo-Hookean, I get this accuracy:
```
min difference analytical : -62.836552257583556 -62.836552256758196
max difference analytical : 154.6409680109583 154.64096800519826
err abs min max: 2.1826984664130578e-13 7.921903133478736e-09
err rel min max: 6.783781664469288e-14 1.7178265731419406e-07
```
so the worst relative error is 1.7e-7. I would like to see similar accuracy with Ogden, when it is set to behave like the neo-Hookean material `(coef=1, exponent=2)`, and I get:
```
min difference analytical : -62.83655226024809 -62.836552256758196
max difference analytical : 154.64096801108747 154.64096800519826
err abs min max: 5.3290705182007514e-14 1.292842455313803e-08
err rel min max: 3.5298307915742917e-15 4.0282612595489903e-07
```
which seems OK. Can you verify my results? The above numbers are with `ogpars=[coef, exponent]`, IMO you have the parameters confused in the example. It might be good to use two parameters explicitely like in dw_tl_he_ogden_ad (which cannot be used here, as the matrix mode does not work with no deformation).

But with `(coef=1, exponent=3)` I get:
```
min difference analytical : -66.98921926018286 -66.98921925597318
max difference analytical : 172.35048001855077 172.35048001355767
err abs min max: 4.814204590530835e-14 2.4408092706325792
err rel min max: 6.291200136269114e-14 8.25259216151863
```
so the worst relative error is 8, which is large. In particular in
```
--- a-r 20 biggest (by r): ir, ic, difference, analytical, adiff, rdiff
   72    29 -2.1101e-01  1.5304e+00  1.74e+00  8.25e+00
  144   146  8.1833e-02 -8.3200e-02  1.65e-01  2.02e+00
  146   108  4.9276e-02  1.4691e-01  9.76e-02  1.98e+00
  146   144  8.1833e-02 -4.6248e-02  1.28e-01  1.57e+00
```
there are even different signs of the values...

However, the Newton solver converges OK (5 iterations per time step instead of 4 for neo-Hook), so the tangent cannot be that bad.

> Should the mesh have been even smaller?

Yes, it would be faster to use just one or two elements (bricks?) for testing.

PR #1056, opened 2023-12-30
L2 constant field

rc:


rc:
Hey @vlukes, what do you think?

vlukes:
Great! I just suggest to add a note about the new field to the User's Guide.

rc:
Good remark, I have added the new file to module index.  The various field kinds are not documented at all, so I will make another PR which will add a field table a la the solver table.

PR #1070, opened 2024-01-31
report status of nonlinear steps

vlukes:
Report also info for each time step. The question is whether to report this info by default.

### DIFF
--- a/sfepy/discrete/problem.py
+++ b/sfepy/discrete/problem.py
@@ -1528,22 +1528,23 @@ class Problem(Struct):
                 for key in time_stats.keys():
                     output('%12s: %.8f [s]' % ('nls ' + key, time_stats[key]))
 
-            step_stats = status.get('step_stats')
-            if step_stats is not None:
-                output('====== step stats ======')
-                if len(step_stats) > 1 and step_stats[0].get("step") > 0:
-                    output(f'step: {1:5} (time=0.0) - initial step')
-                for step in step_stats:
-                    msg = f'step: {step.get("step") + 1:5} '
-                    msg += f'(time={round(step.get("step_time"), 12)}): '
-                    msg += f'cond: {step.get("condition")}, '
-                    msg += f'iter: {step.get("n_iter"):3}, '
-                    msg += f'ls_iter: {step.get("ls_n_iter"):3}, '
-                    msg += f'err0: {step.get("err0"):.3e}, '
-                    msg += f'err: {step.get("err"):.3e}, '
-                    msg += f'time: {step.get("time"):.3f} [s]'
+            if tss.nls.conf.report_status:
+                step_stats = status.get('step_stats')
+                if step_stats is not None:
+                    output('====== step stats ======')
+                    if len(step_stats) > 1 and step_stats[0].get("step") > 0:
+                        output('step: 1 (time=0.0): initial step')
+                    for step in step_stats:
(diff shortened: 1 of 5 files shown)


rc:
Thanks!

What if the number of steps is large? My plan was to redirect the output to a file and then the status lines could be grepped, giving essentially the same information as the output of this PR. So what about making the printed details optional, or to a file? A new option (e.g.` 'convergence_details'` with values `False`, `True` (= print), `'filename'` (= save to the file)) could do the job.

vlukes:
I think the step info is the most important information that users need and this reporting should be enabled by default. What is your opinion?

There is `report_status` option in the solver configuration. Why do not use it also for this?

rc:
Agreed, it is important. But is it practical to have hundreds or thousands of lines printed? The same is partially true also for the output we already have, but that happens on the fly, so one can use it to monitor a solution progress.

Anyway, It would be more practical, if the information was machine readable - what about storing it in a csv file (named like <output-file-name)_log.csv)? Such a file can be read e.g. into pandas using a single line. It might be also preferable to append to that file step by step, so that in case of a calculation failure there are some data to look at.

BTW, I am already using simple scraping of the sfepy output to build such pandas "tables" by employing functions in soops. It is usually tailored to a specific application, so I am in favor of logging the convergence incrementally into a csv file in a unified way. Other information that I usually scrape from the output are the active DOF counts, matrix shape etc. We may think how to store that as well.

vlukes:
> Agreed, it is important. But is it practical to have hundreds or thousands of lines printed? The same is partially true also for the output we already have, but that happens on the fly, so one can use it to monitor a solution progress.
> 

If you know that your problem solution has hundreds of time steps, you can switch the final report off. But for a few time steps it is nice to have a quick convergence check. 

> Anyway, It would be more practical, if the information was machine readable - what about storing it in a csv file (named like <output-file-name)_log.csv)? Such a file can be read e.g. into pandas using a single line. It might be also preferable to append to that file step by step, so that in case of a calculation failure there are some data to look at.

Yes, I agree. Maybe a XML file would be better because you can store more different information in it.

> 
> BTW, I am already using simple scraping of the sfepy output to build such pandas "tables" by employing functions in soops. It is usually tailored to a specific application, so I am in favor of logging the convergence incrementally into a csv file in a unified way. Other information that I usually scrape from the output are the active DOF counts, matrix shape etc. We may think how to store that as well.

sfepy/discrete/problem.py
```diff
@@ -1447,9 +1448,7 @@ def solve(self, state0=None, status=None, force_values=None,
         status : dict-like, optional
             The user-supplied object to hold the solver convergence statistics.
         force_values : dict of floats or float, optional
-            If given, the supplied values override the values of the essential
-            boundary conditions.
-        var_data : dict, optional
+            If given, the supreport_nls_statusonal
```
rc, line 1451: Edited by mistake?

vlukes: Oh, yes, perhaps too sensitive delete key!

rc:
Thanks!

Can you change the log file suffix to `.csv`? It is useful for file action associations (e.g. in `mc` - do you know [visidata](https://www.visidata.org/)?) Also, most functions have the verbose argument as the last - maybe move the new ones so that this holds?

rc:
LGTM!

PR #1073, opened 2024-02-02
move meshes from '2d' to '3d', fix format

vlukes:


vlukes:
Perhaps something wrong in CI tests.

rc:
It happens some times. I do not know the reason for the illegal instruction problem. Let us merge this.

PR #1074, opened 2024-02-03
visualize probes

rc:
It can be tested e.g. with:

```
sfepy-run sfepy/examples/linear_elasticity/its2D_4.py
sfepy-probe sfepy/examples/linear_elasticity/its2D_4.py its2D.h5
sfepy-view its2D.h5 -2 --probes its2D*.txt
```

Currently the annotations are displayed only in the plot 0.

### DIFF
--- a/sfepy/discrete/probes.py
+++ b/sfepy/discrete/probes.py
@@ -2,6 +2,8 @@
 from __future__ import absolute_import
 import hashlib
 
+from ast import literal_eval
+
 import numpy as nm
 import numpy.linalg as nla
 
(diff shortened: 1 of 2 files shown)


vlukes:
I'm thinking how to make the probes more visible. What about adding some labels/legends/...?

rc:
Note that normally the probes are not over the domain edges, where `Variable.get_state_in_region()` can be easily used, so this helps with the visibility. But a label saying each probe's name would be handy.

vlukes (approved):

rc:
Labels added...

vlukes (approved):

PR #1083, opened 2024-02-23
stiffness function for an orthotropic material

vlukes:


sfepy/mechanics/matcoefs.py
```diff
@@ -146,6 +146,51 @@ def youngpoisson_from_stiffness(stiffness, plane='strain'):
 
     return young, poisson
 
+
+def stiffness_from_youngpoissonshear_ortho3(young, poisson, shear):
+    r"""
+    Compute 3D stiffness tensor :math:`{\bm D}` of a orthotropic linear
```
rc, line 152: an orthotropic

sfepy/mechanics/matcoefs.py
```diff
@@ -146,6 +146,51 @@ def youngpoisson_from_stiffness(stiffness, plane='strain'):
 
     return young, poisson
 
+
+def stiffness_from_youngpoissonshear_ortho3(young, poisson, shear):
```
rc, line 150: The name follows the system, but seems too long. What about `stiffness_from_yps_ortho3` or something like that? BTW. What does the "3" mean?

vlukes: "3" indicates that the function is only for 3D problems.

rc:
OK, thanks!

PR #1105, opened 2024-05-23
new octave eigenvalue solver

vlukes:
Use Octave to solve eigenvalue problems. [oct2py](https://pypi.org/project/oct2py/) package is required (can be installed by pip).


rc:
Thanks! Can you add it to `sfepy/tests/test_eigenvalue_solvers.py` ?

rc:
And maybe also add `oct2py` to `check_versions()` in `setup.py`.

rc:
```
AttributeError: module 'sfepy.version' has no attribute 'OCT2PY_IN_VERSION'. Did you mean: 'OCT2PY_MIN_VERSION'
```

sfepy/solvers/eigen.py
```diff
@@ -445,6 +451,56 @@ def __call__(self, mtx_a, mtx_b=None, n_eigs=None, eigenvectors=None,
...
+            'la': (nm.real, -1),
+            'sa': (nm.real, 1),
+            'lr': (nm.real, -1),
+            'sr': (nm.real, 1),
+            'li': (nm.imag, -1),
+            'si': (nm.imag, 1),
+            'be': (nm.real, 1),
+        }
+
+        sfun, order = sort_funs[which.lower()]
```
VascoSch92, line 489: you don't need to use `.lower()` here. 

You are already using lower at line 476.

Same as line 492.

vlukes: Thanks, fixed.

vlukes:
Comparison of solution times of my eigenvalue problem (modal analysis):

-  `eigsh` from Scipy : **2648s !**
- Octave: 214s
- Matlab: 224s

All of these solvers use ARPACK routines.

Solver settings:
```
        'eig': ('eig.scipy', {
             'method': 'eigsh',
             'which': 'LM',
             'sigma': 0.,
             'tol': 1e-3,
             'maxiter': 100, #?
        }),
       'eig': ('eig.octave', {
            'method': 'eigs',
            'which': 'sm',
            'eps': 1e-3,
        }),
       'eig': ('eig.matlab', {
            'method': 'eigs',
            'which': 'sm',
            'eps': 1e-3,
        }),
```

Any idea how to speed-up the Scipy solver?

I found [this discussion](https://stackoverflow.com/questions/59416098/finding-smallest-eigenvectors-of-large-sparse-matrix-over-100x-slower-in-scipy/60042695#60042695) , but I'm not sure if it helps.

vlukes:
- primme solver: 458s

```
        'eig': ('eig.primme', {
           'which': 'SM',
        }),
```

setup.py
```diff
@@ -128,6 +128,8 @@ def check_versions(show_only=False):
                   show_only=show_only)
     package_check('primme', INFO.PRIMME_MIN_VERSION, optional=True,
                   version_getter=_primme_version, show_only=show_only)
+    package_check('oct2py', INFO.OCT2PY_MIN_VERSION, optional=True,
+                  version_getter=_primme_version, show_only=show_only)
```
rc, line 132: The (wrong) `version_getter` should be removed, as `oct2py.__version__` works fine. :)

rc:
@vlukes Not sure why `eig.scipy` is so slow in your case. Maybe send me the problem?

rc (approved):

PR #1109, opened 2024-06-07
sfepy-view: fix vector field check

rc:


vlukes (approved):
It seems to be working.

PR #1113, opened 2024-06-12
flexoelectricity mixed formulation terms

rc:


sfepy/terms/terms_flexo.py
```diff
@@ -0,0 +1,190 @@
+"""
+Flexoelectricity related terms.
+"""
+import numpy as nm
+
+from sfepy.mechanics.tensors import dim2sym
+from sfepy.terms.terms_multilinear import ETermBase
+
+def make_grad2strain(dim):
```
VascoSch92, line 9: what happens if `dim` is not 1, 2, or 3?

sfepy/terms/terms_flexo.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def get_function(self, mat, tvar, svar, mode=None, term_mode=None,
+                     diff_var=None, **kwargs):
+        dim = self.region.dim
+        sym = dim2sym(dim)
+        aux = make_grad2strain(dim)
+        n_qp = mat.shape[1]
+        mat = mat.reshape((-1, n_qp, dim, dim, sym))
+
+        if term_mode is None:
```
VascoSch92, line 133: what if `term_mode` is not one of 'electric_displacement',  'double_stress' or `None` ?

sfepy/terms/terms_flexo.py
rc: Thanks, fixed!

sfepy/terms/terms_flexo.py
rc: Thanks, fixed!

rc:
The CI failures are due to numpy 2.0.0 - I forgot to pin the version to <2.

PR #1119, opened 2024-06-24
add more meshio formats to supported formats lists

rc:
@vlukes, maybe we could define another flag, for example 'R' to denote that the format is read-only (so that 'w' is not automatically added)?


vlukes:
@rc, do you have any tests for the newly enabled formats? 

> @vlukes, maybe we could define another flag, for example 'R' to denote that the format is read-only (so that 'w' is not automatically added)?

Do you mean 'R' to be a special flag to override meshio automatic detection?

rc:
> @rc, do you have any tests for the newly enabled formats?
> 
> > @vlukes, maybe we could define another flag, for example 'R' to denote that the format is read-only (so that 'w' is not automatically added)?
> 
> Do you mean 'R' to be a special flag to override meshio automatic detection?

Yes, so that the 'w' is not automatically added.

rc:
OK, let's merge it as it is...

PR #1132, opened 2024-08-02
contact term based on IPC toolkit

rc:
This PR introduces a new term for contacts based on [IPC toolkit](https://github.com/ipc-sim/ipc-toolkit).

Example run:
```
sfepy-run sfepy/examples/linear_elasticity/two_bodies_contact.py -d "dims0=(1.0, 1.0, 0.5),shape0=(5, 5, 4),centre0=(0, 0, -0.25),dims1=(1.2, 0.8, 0.5),shape1=(9, 9, 6),centre1=(0, 0, 0.25),shift10=(0, 0, 1e-4),shift11=(0, 0, -0.2), n_step=5,contact='ipc',output_dir='output/ipc'"

python3 sfepy/scripts/plot_logs.py output/ipc/log.txt

sfepy-view output/ipc/two_bodies.h5 -f u:wu:f1:p0 1:vw:wu:f1:p0
```


zfergus:
Hi @rc, 

* Yes, we use IPC for quasi-static problems. A differences is that without the inertia term/mass matrix in the optimization, it is possible the system is under-determined. You should constrain at least three vertices (two in 2D) to remove the rotational null space. This can either be done per free body or for a single body given that all other bodies are in contact. The latter can sometimes be tricky to guarantee as bodies move and slide out of contact.
* Yes, the IPC Toolkit is capable of handling 2D objects and we use it for this case. You should pass a Nx2 matrix for the vertex positions to `CollisionMesh`. In this case the only contacting geometry are the boundary edges, so there is no need to pass the interior triangles to the `CollisionMesh`.

Hope this helps,
Zach

rc:
> * Yes, the IPC Toolkit is capable of handling 2D objects and we use it for this case. You should pass a Nx2 matrix for the vertex positions to `CollisionMesh`. In this case the only contacting geometry are the boundary edges, so there is no need to pass the interior triangles to the `CollisionMesh`.

Yes! I was passing `faces=None` and got
```
    collision_mesh = self.ipc.CollisionMesh(smesh.coors, edges, faces)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. ipctk.CollisionMesh(rest_positions: numpy.ndarray[numpy.float64[m, n]], edges: numpy.ndarray[numpy.int32[m, n]] = array([], shape=(0, 0), dtype=int32), faces: numpy.ndarray[numpy.int32[m, n]] = array([], shape=(0, 0), dtype=int32), displacement_map: scipy.sparse.csc_matrix[numpy.float64] = <Compressed Sparse Column sparse matrix of dtype 'float64' with 0 stored elements and shape (0, 0)>)
    2. ipctk.CollisionMesh(include_vertex: list[bool], full_rest_positions: numpy.ndarray[numpy.float64[m, n]], edges: numpy.ndarray[numpy.int32[m, n]] = array([], shape=(0, 0), dtype=int32), faces: numpy.ndarray[numpy.int32[m, n]] = array([], shape=(0, 0), dtype=int32), displacement_map: scipy.sparse.csc_matrix[numpy.float64] = <Compressed Sparse Column sparse matrix of dtype 'float64' with 0 stored elements and shape (0, 0)>)
```

Omitting the `faces` argument works as you said. Looking at the error message I did not get where is the problem.

Thank you for confirming that the quasi-static case should work. I will debug some more and probably return with further questions.

zfergus:
I think the problem with passing None is pybind11 strictly enforces the type, so it has to be a numpy array. The array can be empty (default) for 2D.

rc:
More questions are here, I apologize for the length, but I wanted to explain
properly my ipctk use.

SfePy builds equations from independent composable terms, so I had to make some
adjustments for the need of the new ipctk-based term to access the elastic
energy gradient and other information. I have also modified the Newton solver
to allow custom line search functions. See below, but it is a bit long :(.

My example involves two rectangular boxes, discretized by hexahedral elements.
The bottom one is completely fixed on the bottom face, while the top one's top
face moves down (lateral DOFs are fixed, so rigid body motions should not spoil
the elasticity Hessian)) as time progresses in 5 quasistatic time steps. Both
boxes use the linear elasticity formulation for simplicity, even though the
displacements/deformations are not small. In a consistent unit set, the default
dimensions and material parameters are:
```
dims0=(1.0, 1.0, 0.5), # Box 0 dimensions.

dims1=(1.2, 0.8, 0.5), # Box 1 dimensions.

young=1.0, # Young's modulus.
poisson=0.3, # Poisson's ratio.
rho=1.0, # Density.

cm=None, # IPC average mass, None means calculated as domain volume * rho.
ck=0.0, # IPC barrier stiffness, 0.0 means adaptive.
dhat=1e-2, # IPC barrier activation distance.
pspd='NONE', # Hessian projection
```

I have put several runs and their results into the attached Jupyter noteboook [1].
Each run is documented by a table of parameters, Newton convergence log plot
(time steps marked by vertical lines), contact log plot (barrier distance,
stiffness, elastic and barrier residual norms) and final time deformation plot.

[1] [ipctk-tests.ipynb.gz](https://github.com/user-attachments/files/17097822/ipctk-tests.ipynb.gz)

The results are as follows:

- With the default parameters, specifically no Hessian projection (see the
  notebook), the Newton solver seems to converge OK.

- With the 'CLAMP' projection, the second time step convergence stalls.

- With the dimensions and parameters adjusted for units: `mm, ms, kg -> kN, GPa, J, kg/mm^3 = 1e9 kg/m^3`, no Hessian projection, the convergence is slow in all time steps after the first one and stalls in the last.

- With same as above, 'CLAMP' Hessian projection, also the convergence is similar to the above case, but the distance function oscillates less (FWIW).

IMO I am doing something very wrongly to have such a bad convergence for this
very small example.

For reference, what the implementation currently does can be expressed by the
following Python-like pseudocode:

```python
def newton(f, u0):
    """
    Newton method with modified line-search for solving f(u) = 0, initial
    guess is u0.
    """
    u = u0
    du = 0
    it = 0
    err0 = 0
    while 1:
        u, r, err = apply_line_search(f, u0, du, it, err0)
        if check_convergence(r, r0, err):
            break

        A = f(u, mode='matrix')

        du = solve_linear(A, r)
        err0 = err
        it += 1

    return u

def apply_line_search(f, u0, du0, it, err0):
    ls = 1.0 # backtracking line search factor.
    du = du0
    ls_it = 0
    while 1:
        u = u0 - du
        if it > 0: # du = 0 when it == 0.
            # mesh.coors ... undeformed mesh coordinates.
            # nods ... indices of contact nodes.
            x0 = mesh.coors[ic] + u0[nods]
            x1 = mesh.coors[ic] + u[nods]

            max_step_size = ipctk.compute_collision_free_stepsize(
                collision_mesh, x0, x1,
            )

            if max_step_size < 1.0:
                u = u0 - max_step_size * du

            ls = min(ls, max_step_size)

        r1 = f(u, 'omit_contact_term') # elastic energy gradient
        r2 = f(u, it=it, ls_it=ls_it, e_grad=r1, 'contact_term_only)')
        r = r1 + r2

        err = |r|
        if err < err0:
            break

        ls /= 0.5
        du = ls * du0
        ls_it += 1

    return u, r, err

def f_contact_term(u, it, ls_it, e_grad, mode='vector'):
    """
    During the first call, initialize contact info with:

    - collision_mesh - created from
       - surface faces if simplex mesh
       - triangulated surface faces if tensor product mesh

    - nods - indices of contact region nodes in the whole mesh

    Term parameters:
    - dhat
    - avg_mass - volume * density
    - Pspd - SPD projection mode
    """
    vertices = collision_mesh.rest_positions + u[nods]
    collisions = ipctk.Collisions()
    collisions.build(collision_mesh, vertices, dhat)

    ci.min_distance = collisions.compute_minimum_distance(
        collision_mesh, vertices,
    )
    ci.bbox_diagonal = ipctk.world_bbox_diagonal_length(vertices)

    B = ipctk.BarrierPotential(dhat)

    # Adaptive stiffness computation.
    if it == 0: # time step start - 1. Newton iteration.
        bp_grad = B.gradient(collisions, collision_mesh, vertices)

        (ci.barrier_stiffness,
         ci.max_barrier_stiffness) = ipctk.initial_barrier_stiffness(
             bbox_diagonal,
             B.barrier,
             dhat, avg_mass,
             e_grad, bp_grad,
         )
        actual_stiffness = ci.barrier_stiffness
        ci.prev_min_distance = ci.min_distance

    elif (ls_it == 0) and (mode == 'vector'):
        # Do not update stiffness during line-search.
        actual_stiffness = ipctk.update_barrier_stiffness(
            ci.prev_min_distance, ci.min_distance,
            ci.max_barrier_stiffness, ci.barrier_stiffness,
            ci.bbox_diagonal,
        )
        ci.barrier_stiffness = actual_stiffness
        ci.prev_min_distance = ci.min_distance

    else: # matrix mode or within line-search.
        actual_stiffness = ci.barrier_stiffness

    if mode == 'vector':
        bp_grad = B.gradient(collisions, collision_mesh, vertices)

        out = actual_stiffness * bp_grad

    else: # mode == 'matrix'
        # Note: The elastic part of the Hessian is not projected.
        bp_hess = B.hessian(
            collisions, collision_mesh, vertices,
            project_hessian_to_psd=Pspd,
        )
        out = actual_stiffness * bp_hess

    return out
```

rc:
I will merge this now, as the example works, and I did not have time to improve it more.

PR #1139, opened 2024-08-20
new surface term (term name?)

vlukes:
I think we don't have a term that does the same job as the one in this PR. @rc, do you have an idea how to name this surface term?

### DIFF
--- a/sfepy/terms/terms_multilinear.py
+++ b/sfepy/terms/terms_multilinear.py
@@ -1436,6 +1436,45 @@ class EDivGradTerm(ETermBase):
 
         return fun
 
+
+class ESurfXTerm(ETermBase):
+    r"""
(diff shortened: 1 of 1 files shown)


rc:
Yes, it is not available. What is the context of its use? It seems like a flux but it is tested by a vector variable,

vlukes:
Surface traction in Stokes flow:

\int \ul{v} \cdot (\ull{\sigma} cdot \ul{n})

where

\sigma = - p I + \nu \nabla \ul{u}

If p is given, than the second term (new one in PR) will appear on the LHS.

rc:
So what about `StokesTractionTerm` and `de_stokes_traction`? Your explanation above could be added to the docstring.

rc:
Thanks, LGTM!

PR #1148, opened 2024-09-24
migrate to python-mumps interface

vlukes:
The Schur complement method is not working yet. Implementation is in progress.


sfepy/solvers/ls.py
```diff
@@ -842,100 +832,45 @@ def presolve(self, mtx, use_mtx_digest=True):
...
-                self.mumps_ls = self.mumps.MumpsSolver(system=system,
-                                                       is_sym=is_sym,
-                                                       mem_relax=mem_relax)
-            if self.conf.verbose:
-                self.mumps_ls.set_verbose()
-
-            self.mumps_ls.set_mtx_centralized(mtx)
-            self.mumps_ls(4)  # analyze + factorize
+                maxval = mtx.diagonal().max()
+                is_sym = nm.all(nm.abs((mtx - mtx.T).data / maxval) < 1e-9)
```
rc, line 837: Is `is_sym` used anywhere?

rc:
Great, thanks! Did you check the efficiency in comparison with the original wrappers?

rc:
I have yet to check on windows, but on linux it works! How did you install python-mumps? I have used conda+miniforge, because pip failed. Maybe the original wrappers could be kept as well, as python-mumps installation outside conda is not easy? FWIW the original wrappers seem faster, but it is not a good comparison due to different environments.

vlukes:
I have used `pip` to install `python-mumps`.

I can rename the original wrapper to `mumps-o` for now and remove it later. OK?

rc:
BTW I found the pip error cause - on my system there is no `python` command, only `python3`, so when meson runs https://gitlab.kwant-project.org/kwant/python-mumps/-/blob/main/mumps/_version.py, it fails. Fixed by a symlink :)

The new solver is slower. I have edited `sfepy/tests/test_linear_solvers.py` to refine the mesh once and commented out other solvers. Then:
 - original:
```
sfepy-test -vs sfepy/tests/test_linear_solvers.py -k test_solvers
...
... 2.69 [s] /   -1 (1.103e-14) : d10 ls.mumps
```
- new:
```
... 3.46 [s] /   -1 (1.117e-14) : d10 ls.mumps
```

For the original small mesh it was even more pronounced. So I would definitely keep the original wappers too. And maybe support both in the auto-fallback solver, with the original having the highest priority?

vlukes:
I am closing this PR due to the new implementation in #1153.

PR #1153, opened 2024-10-03
External wrappers for MUMPS solver

vlukes:
This PR modifies the `ls.mumps` solver such that it tries to import [mumpspy](https://pypi.org/project/mumpspy) wrapper at first. If the import fails, it will try to import [python-mumps](https://pypi.org/project/python-mumps/) instead. Now only `mumpspy` can solve a linear system using the Schur complement method. The PR for the complete Schur solution has been sent to the `python-mumps` project. If it is merged, I will modify the `ls.mumps_schur` solver.

The new PR replaces #1148.


rc:
Nice, thanks! I will check the timings tonight. :)

BTW. what about transferring mumpsy under the sfepy github org? Just an idea...

rc:
In contrast with #1148, now the solution speed seems the same with mumpspy or python-mumps.

rc:
I have a small fix for no mumps interface installed - can I push here?

vlukes:
> I have a small fix for no mumps interface installed - can I push here?

Yes, push it here.

rc:
Are we ready to merge?

vlukes:
> Are we ready to merge?

Could you wait a while? I'm finishing the implementation of the complete Schur solution in the python-mumpy package.

rc:
Of course, no hurry...

sfepy/solvers/ls.py
```diff
@@ -802,15 +802,38 @@ class MUMPSSolver(LinearSolver):
...
+            for module in modules:
+                try:
+                    return importlib.import_module(module), module
+                except ImportError:
+                    continue
 
+            msg = 'cannot import MUMPS! Install either mumspy or python-mumps.'
+            raise ImportError(msg)
+
+        aux = try_imports(['import mumpspy as mumps', 'import mumps'],
```
rc, line 832: So the `mumps_import()` function above is not needed anymore?

PR #1155, opened 2024-10-10
debug warped cells site configuration option

rc:


rc:
@vlukes min_qp(J) per cell is now saved too.

PR #1164, opened 2024-12-05
fix multi-point constraints weights handling

rc:
The weights can actually be tried now.


vlukes (approved):

vlukes:
Great! The results are close to those reported in [1] even for non-unit weights.

[1] http://web.mscsoftware.com/support/online_ex/previous_nastran/nas105/workshop_10.pdf

rc:
OK, thanks for verifying!

PR #1181, opened 2025-02-07
improve corretors and get_homog_coefs_nonlinear()

vlukes:
* new: dump actual corr. name and component to problem.homogen_corr_id
* get_homog_coefs_nonlinear(): return correctors if required
* remove dependency on six


rc:
Thanks! Where is `problem.homogen_corr_id` actually used? Just curious :)

vlukes:
I use `problem.homogen_corr_id` in `scale_solution_fun()` and  `scale_system_fun()` of the nonlinear solver, where I need to know which corrector and its component is currently being solved. This is a part of the model order reduction algorithm.

PR #1184, opened 2025-02-11
PyPardiso linear solver

rc:


vlukes (approved):

rc:
PyPardiso seems to be faster than MUMPS (but maybe more memory demanding, needs more tests) - what about adding it to `AutoDirect` as the first choice?

rc:
Maybe later, after it gets some testing...

vlukes:
Yes, I agree.

PR #1185, opened 2025-02-12
New general hyperelastic term

vlukes:
New "general" HyperelasticUL term.

The tangent modulus and stress tensor are evaluated by the user in a function specified in `materials` :
```
materials = {
    'solid': ({'.fun': hyperelastic_mat_function},),
}

equations = {
    'balance': 'dw_ul_he_by_fun.i.Omega(solid.fun, v, u) = 0',
}
```

Function `hyperelastic_mat_function` is defined as follows:
```
def hyperelastic_mat_function(family_data, mode):
    ...
    return out
```
where `mode` can be either `tan_mod` or `stress`.

### DIFF
--- a/sfepy/examples/large_deformation/hyperelastic_ul_by_fun.py
+++ b/sfepy/examples/large_deformation/hyperelastic_ul_by_fun.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+r"""
+Compressible Neo-Hookean hyperelastic material model. The tangent modulus
+and the stress tensor are calculated by a user defined function.
+"""
+import numpy as nm
+from sfepy import data_dir
+from hyperelastic_ul import (filename_mesh, options, regions, fields,
+    variables, ebcs, functions)
+
+mu, K = 20., 1000.
+
+# function which returns tangent modulus and stress tensor
+def get_hyperelastic_mat(family_data, mode):
+    from sfepy.terms.extmods.terms import sym2nonsym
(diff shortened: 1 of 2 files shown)


rc:
Thanks! Do you plan to add also a TL term?

sfepy/examples/large_deformation/hyperelastic_ul_by_fun.py
```diff
@@ -0,0 +1,102 @@
...
+        'i_max': 20,
+        'eps_a': 1e-5,
+        'eps_r': 1e-3,
+        }),
+    'ts': ('ts.simple', {
+        't0': 0,
+        't1': 1,
+        'n_step': 11,
+        }),
+    }
```
rc, line 102: Note: different indentation style then used with other keywords.

sfepy/examples/large_deformation/hyperelastic_ul_by_fun.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
+r"""
+Compressible Neo-Hookean hyperelastic material model. The tangent modulus
+and the stress tensor are calculated by a user defined function.
+"""
+import numpy as nm
+from sfepy import data_dir
```
rc, line 7: `data_dir` is not used.

vlukes:
No, I don't currently plan to do that. I need the term for nonlinear 
two-scale homogenization which is formulated in the ULF.

rc:
The tests pass, but maybe we should check the shape of `mat_fun()` return values to warn user when it is wrong...

PR #1186, opened 2025-02-18
assemble matrices with long wors using a binary search

rc:
This significantly speeds up assembling with L2 constant fields.

### DIFF
--- a/sfepy/discrete/common/extmods/assemble.pyx
+++ b/sfepy/discrete/common/extmods/assemble.pyx
@@ -187,3 +187,131 @@ def assemble_matrix_complex(np.ndarray[complex128, mode='c', ndim=1]
                 else:
                     msg = 'matrix item (%d, %d) does not exist!' % (irg, icg)
                     raise IndexError(msg)
+
+cdef inline int32 bsearch(int32 *cols,
+                          int32 i0,
+                          int32 i1,
+                          int32 ii) nogil:
+    cdef int32 im
+
+    while i0 <= i1:
+        im = (i0 + i1) // 2
+
+        if cols[im] == ii:
+            return im
+
+        elif cols[im] < ii:
+            i0 = im + 1
+
+        else:
+            i1 = im - 1
+
+    return -1
+
+@cython.boundscheck(False)
+def assemble_matrix_b(np.ndarray[float64, mode='c', ndim=1] mtx not None,
+                      np.ndarray[int32, mode='c', ndim=1] prows not None,
+                      np.ndarray[int32, mode='c', ndim=1] cols not None,
+                      np.ndarray[float64, mode='c', ndim=4] mtx_in_els not None,
+                      np.ndarray[int32, mode='c', ndim=1] iels not None,
+                      float64 sign,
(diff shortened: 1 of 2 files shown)


sfepy/discrete/common/extmods/assemble.pyx
```diff
@@ -187,3 +187,131 @@ def assemble_matrix_complex(np.ndarray[complex128, mode='c', ndim=1]
...
                     raise IndexError(msg)
+
+cdef inline int32 bsearch(int32 *cols,
+                          int32 i0,
+                          int32 i1,
+                          int32 ii) nogil:
+    cdef int32 im
+
+    while i0 <= i1:
+        im = (i0 + i1) // 2
```
VascoSch92, line 198: There is a potential integer overflow issue here.

If `i0` and `i1` are both large positive values, their sum `i0` + `i1` can exceed $2^{32} - 1$, leading to integer overflow.

A solution is to rewrite it with `im = i0 + (i1 - i0) // 2`.

rc: Would `im = <int32> ((<uint32> i0 + <uint32> i1) >> 1)` work as well? It has fewer operations, and I need this to be as fast as possible.

VascoSch92: I don't know... I will try to run some tests and see what happens.

However, I don't think that` i0 + (i1 - i0) // 2` will slow down the speed of the program excessively. Do you have some benchmark?

rc: I have just measured the `bsearch()` using those lines on a small matrix. In 1152 calls the retyping takes about 4.1657e-05 s, while the subtraction about 4.3355e-05 s. But that is just from two measurements using:
```
from posix.time cimport clock_gettime, timespec, CLOCK_MONOTONIC

cdef inline int32 bsearch(int32 *cols,
                          int32 i0,
                          int32 i1,
                          int32 ii):
    cdef int32 im
    cdef timespec start, end
    cdef double elapsed_time

    clock_gettime(CLOCK_MONOTONIC, &start)
    while i0 <= i1:
        # im = <int32> ((<uint32> i0 + <uint32> i1) >> 1)
        im = i0 + ((i1 - i0) >> 1)
        # im = (i0 + i1) >> 1

        if cols[im] == ii:
            clock_gettime(CLOCK_MONOTONIC, &end)

            # Calculate elapsed time in seconds
            elapsed_time = (end.tv_sec - start.tv_sec) + (end.tv_nsec - start.tv_nsec) / 1e9
            print(elapsed_time)
            return im

        elif cols[im] < ii:
            i0 = im + 1

        else:
            i1 = im - 1

    return -1
```
and then summing up all printed times.

rc: I will go with the retyping version. As a matter of fact, the whole C/Cython types would need to be revisited for very large problems/matrices.

rc: And thanks for the review!

PR #1187, opened 2025-02-20
new CoefSymDim

rc:


vlukes (commented):
`CoefDimSym` already exists. Could not be used transapposition?

rc:
I did not realize that the order in `'set_variables'` matters. So yes, `CoefDimSym` could be used with `CoefEval` for the transposition, if needed. This seems easier, though.

PR #1189, opened 2025-03-07
clean up homogenized coefficients

rc:
Remove obsolete/unused code.


rc:
@vlukes, what about `coefs_perfusion.py`? It is also unused in sfepy examples and the code seems bit-rotten.

vlukes:
> @vlukes, what about `coefs_perfusion.py`? It is also unused in sfepy examples and the code seems bit-rotten.

Remove it.

vlukes (approved):

PR #1191, opened 2025-03-11
create matrix graph using COO matrix

rc:
No C, and faster for large meshes than the original. TODO: make the chunk size configurable.


sfepy/discrete/fem/mesh.py
```diff
@@ -536,14 +536,14 @@ def create_conn_graph(self, verbose=True):
...
-        output('graph nonzeros: %d (%.2e%% fill)' \
-               % (nnz, 100.0 * float(nnz) / nm.prod(shape)), verbose=verbose)
+        graph = 0
+        for tdim in self.dims:
+            conn = self.get_conn(self.descs[0], tdim=tdim)
+            graph += create_dof_graph(conn, conn, shape).tocsr()
 
-        data = nm.ones((nnz,), dtype=bool)
-        graph = sp.csr_matrix((data, icol, prow), shape)
+        output('...done in %.2f s' % timer.stop(), verbose=verbose)
```
vlukes, line 544: I would use f-strings if possible.

vlukes (approved):
Great! Many lines of C code disappear.

sfepy/discrete/fem/mesh.py
rc: Yeah, I left those lines as they were. Lots of places would need such a facelift.

PR #1196, opened 2025-03-13
fix scalar range in pv_plot() for non-vectors (tensors)

rc:
Fix also scalar bar label, as in '|text|' the vertical bars are ignored by pyvista.

### DIFF
--- a/sfepy/scripts/resview.py
+++ b/sfepy/scripts/resview.py
@@ -591,6 +591,9 @@ def pv_plot(filenames, options, plotter=None, step=None, annotations=None,
                 plot_info.append('glyphs=grad(%s), factor=%.2e'
                                  % (field, factor))
 
+                scalar = 'gradient Magnitude'
+                scalar_label = f'grad({field}):mag'
+
         elif 'c' in opts and has_components:  # select field component
             comp = opts['c']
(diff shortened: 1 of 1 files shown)


sfepy/scripts/resview.py
```diff
@@ -546,19 +545,25 @@ def pv_plot(filenames, options, plotter=None, step=None, annotations=None,
         if opts.get('l', options.outline):  # outline
             plotter.add_mesh(pipe[-1].outline(), color='k')
 
-        scalar = field
-        scalar_label = scalar
         is_vector_field = ((field is not None)
```
vlukes, line 548: This can be moved into `if 'g' in opts and is_point_field` block.

vlukes (approved):

sfepy/scripts/resview.py
rc: is_vector_field is used also in streamlines?

rc:
Now for a scalar field with 'g', the scalar bar will show the scalar, not the gradient magnitude. I will change that, OK?

vlukes:
> Now for a scalar field with 'g', the scalar bar will show the scalar, not the gradient magnitude. I will change that, OK?

OK

PR #1204, opened 2025-04-02
update for numpy 2.0

rc:


sfepy/base/compat.py
```diff
@@ -8,10 +8,14 @@
...
 
 except AttributeError:
-    in1d = nm.setmember1d
+    try:
+        in1d = nm.in1d
+
+    except AttributeError:
+        in1d = nm.setmember1d
 
 unique = nm.unique
```
vlukes, line 20: Import of `nm.unique1d` could also be removed. Couldn't it?

vlukes (approved):

sfepy/base/compat.py
rc: Ir is imported only in `compat.py`, so that the code still works with older numpy versions.

rc: But it seems unused, so yes, it can be removed.

PR #1228, opened 2025-08-19
New keyword "full_order" in integral configuration

vlukes:


rc:
Thanks! There is a debug output that could be removed, and also the [83cdfd9](https://github.com/sfepy/sfepy/pull/1228/commits/83cdfd9302684df7a7e7bf88a2543c3731c5d8d4) is already merged from #1227.

PR #1232, opened 2025-10-04
trim Python 2 hybridation

a-detiste:
Hi,

It was stated in `INSTALL` in 2020 that only Python 3 is supported.

It will file several easy-to-review, topical PR to remove Python2 review.

Greetings


rc:
Thank you, merging...

PR #1233, opened 2025-10-04
replace "basestr" with plain "str" everywhere

a-detiste:
Hi,

Should these two declaration go too ? or maybe some third party rely on these ?

```
sfepy/base/base.py-# is this part of some public API ?
sfepy/base/base.py-PY3 = True
sfepy/base/base.py:basestr = str
```

We are in 2025, I guess no ...

### DIFF
--- a/sfepy/base/base.py
+++ b/sfepy/base/base.py
@@ -19,10 +19,6 @@ sfepy_config_dir = os.path.expanduser('~/.sfepy')
 if not os.path.exists(sfepy_config_dir):
     os.makedirs(sfepy_config_dir)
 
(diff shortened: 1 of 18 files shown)


rc:
> Hi,
> 
> Should these two declaration go to ? or maybe some third party rely on these ?
> 
> ```
> sfepy/base/base.py-# is this part of some public API ?
> sfepy/base/base.py-PY3 = True
> sfepy/base/base.py:basestr = str
> ```
> 
> We are in 2025, I guess no ...

Agreed, you can remove the definitions. Thanks!

PR #1236, opened 2025-10-06
compile with GCC-15

a-detiste:
This is an uptreamable version of this:

https://sources.debian.org/src/sfepy/2025.3-1/debian/patches/gcc_15


rc:
Thank you again!

Issue #1041, opened 2023-11-06
2D axisymmetric FEA / longitudinal cross section of a pipe


megaprimatus:
This is not actually an issue is more like a question so you can close it write away it it dosen't coralete with a policy.

I didn't find any example of the 2D axisymmetric analysis in SFEPY docs. Is there a capability to do this simulation described below on the image? The pipe cross section is under internal pressure simultaneously under tension force.
![2d_axisymmetric](https://github.com/sfepy/sfepy/assets/48415379/fd580c0d-e24c-4b5d-b5ec-bc9ddc872438)

rc:
There are no special axisymmetric elements in SfePy, so it depends on your equations - maybe those can be built using the [available terms](http://sfepy.org/doc-devel/terms_overview.html).

PR #1246, opened 2026-01-05
Add 2D axisymmetric linear elasticity example

zhangrongchuan:
This PR adds a 2D axisymmetric linear elasticity example together with
a derivation document.

Files added:
- sfepy/examples/linear_elasticity/elastic2D_axisymmetric.py
- sfepy/examples/doc/elastic2D_axisymmetric.md

This PR addresses issue #1041.


rc:
I have tried to fix the `m_hoop_idx.val` parameter. I set it to 1 to make the gradient w.r.t. z - is that correct?

PR #1277, opened 2026-04-15
WIP: Nonlinear Parametric Diffusion ; initial version, feedback welcome

ansh-shrivastava-sdsu:
Adds an example combining nonlinear diffusion with a parametric study.

Solves -div((1 + alpha u^2) grad u) = sin(pi x) sin(pi y)
with Dirichlet boundary conditions, for several alpha values.

WIP: feedback welcome.

Thank you!

### DIFF
--- a/sfepy/examples/diffusion/poisson_nonlinear_parametric.py
+++ b/sfepy/examples/diffusion/poisson_nonlinear_parametric.py
@@ -1,16 +1,26 @@
 r"""
 Nonlinear diffusion with a field-dependent coefficient and parametric sweep.
+
 Solve the problem:
+
 .. math::
+
    -\nabla \cdot \left( (1 + \alpha u^2)\nabla u \right)
    = \sin(\pi x)\sin(\pi y)
    \quad \text{in } \Omega,
+
 with homogeneous Dirichlet boundary conditions:
+
 .. math::
+
    u = 0 \quad \text{on } \partial \Omega.
+
 The diffusion coefficient depends on the solution:
+
 .. math::
+
    c(u) = 1 + \alpha u^2.
+
 The problem is solved for several values of :math:`\alpha`, and the
 solutions are compared to the linear case :math:`\alpha = 0`.
 """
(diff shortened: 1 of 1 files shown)


rc:
Thank you for the proposal! It runs, so it is a good beginning :) A few tips for improvement:

- Only the last result file is present, overwriting the previous ones. To save all, you need to change the output file name for each alpha. Check `problem.setup_output()` as used in [this example](https://sfepy.org/doc-devel/examples/diffusion-poisson_parametric_study.html).
- You may want to use the define() function, so that the alphas and maybe other parameters can be easily changed from command line without editing the file.
- The math expressions in the docstring might need empty lines around to render correctly.

ansh-shrivastava-sdsu:
Hi Robert and SfePy Community! 
I fixed the issues that were brought up from my first draft, and believe it's good to go now.

Changes : 
1. Each run saves to a unique output file 
2. Used define() so parameters can be set from the command line 
3. Modified docstring formatting with extra lines for rendering 

The example runs correctly on my end, please let me know if there’s anything else I should improve :) 

Thank you! 
Ansh Shrivastava

rc:
Hi Ansh,

it looks good! Just a few polishing touches:

- The default alphas (while realistic?) lead to no discernible differences when displayed by `sfepy-view`. Wouls using e.g. `alphas=[0, 1000, 10000, 100000, 1000000]` make sense?
- Use something like `alpha_tag = f'{alpha:010.2f}'.replace('.', '_')` so that the files are sorted lexicographically by alpha magnitude.
- Increase the integral order to at least 4 as the material function is non-linear (see also [FAQ](https://sfepy.org/doc-devel/faq.html#finite-element-approximation-field-order-and-numerical-quadrature-order)). maybe add `qp_order` as `define()` parameter, together with `order` for the field approximation order. 
- Further possible parameters for `define()`: `i_max=20, output_dir='output/poisson_nonlinear_parametric'`.

Thanks!

ansh-shrivastava-sdsu:
Hi Robert and SfePy Community!

I implemented the four suggested changes, please let me know if this is good to go. I reran the example successfully, and it now produces separate output files for each parameter value.

@rc 

Thank you,
Ansh Shrivastava

rc:
Thank you, looks good to me. Last little thing to fix: my editor shows some trailing whitespace (spaces at the ends of some lines) in the file. Remove it, please, and then I think it will be ready. Many editors/IDEs can be setup to automatically remove trailing white space, for example on save or on commit.

rc:
I can fix that for you, if you want. (Would need to push to your branch.)

ansh-shrivastava-sdsu:
Hi Robert, sorry for the late reply - fixed the trailing whitespaces. 

Thank you! 
Ansh Shrivastava

rc:
OK, thank you, merging (after CI tests pass)!

